PR #23, opened 2018-04-15
Add Carthage support

mluisbrown:
In order to add support for Carthage, and maintain support for CocoaPods the project had to be completely [re-structured](https://www.quora.com/How-should-an-OSS-library-be-structured-for-distribution-with-both-Carthage-and-CocoaPods/answer/Justin-Spahr-Summers) to not use CocoaPods for sub-dependencies. 

So, this PR does the following:
- De-integrates CocoaPods from the workspace
- Renames the `.xcworkspace` and `.xcproject` files from `Example` to `Bento`
- Adds `Cartfile` and `Cartfile.private` for Carthage support
- Uses git submodules for sub-dependencies (`carthage update --use-submodules`) instead of CocoaPods
- Adds the dependency's Xcode projects to the workspace (from their location in `Carthage/Checkouts`)
- Adds the required framework dependencies to each target as required
- Removes the Podfile
- Updates CI to reflect above changes

I've tested pulling Bento as a dependency in a test project using both Carthage (`github "Babylonpartners/Bento" "michael/carthage"`) and CocoaPods (`pod 'Bento', :git => 'git@github.com:Babylonpartners/Bento.git', :branch => 'michael/carthage'`) and both worked fine.


sergdort:
👍  

I think we need to add a note to the ready how to setup project to run the example. As I believe now we need to run `carthage checkout` right?

mluisbrown:
@sergdort you don't need to run `carthage checkout`. You can just do:

`git submodule update --init --recursive`

mluisbrown:
@sergdort I added a section about Dev Installation.

PR #24, opened 2018-04-15
English fixes

diegopetrucci:


README.md
```diff
@@ -45,10 +45,10 @@ tableView.render(box)
 
 #### Sections and Nodes 🏗
 
-A `Section` and a `Node` are building blocks of the `Box`:
+`Section`s and `Node`s are building blocks of a `Box`:
 
-- The `Section` is an abstraction of `UITableView`'s section, which defines whether there is going to be any header or footer.
-- The `Node` is an abstraction of `UITableView`'s row, it defines how it's going be rendered.
+- `Section` is an abstraction of `UITableView`'s section which defines whether `header`s or `footer` should be shown.
```
sergdort: Should probably be just `header`

README.md
```diff
@@ -9,7 +9,7 @@
 - **Diffing:** reloads your UI with beautiful animations when your data changes
 - **Component-based:**  Design reusable components and share your custom UI across multiple screens of your app
 
-In our experience it makes UI-related code easier to build and maintain. Our aim is to make the UI a function of state (`UI = f(state)`), which makes `Bento` a perfect fit for Reactive Programming.
+In our experience it makes UI-related code easier to build and maintain. Our aim is to make the UI a function of state (i.e: `UI = f(state)`), which makes `Bento` a perfect fit for Reactive Programming.
```
RuiAAPeres: I disagree with the `i.e:` because it's not an example, it's just a different way of saying `Our aim is to make the UI a function of state`.

README.md
mluisbrown: `which defines whether a `header` or `footer` should be shown.`

README.md
mluisbrown: That's what `i.e:` means: Latin `id est`, meaning `that is`, meaning "a different way of saying"

diegopetrucci: To remember what it means I always use "eg = example given, ie = in essence" which is not correct but it fits :)

RuiAAPeres: I mistakenly read it as example ` exempli gratia`/`eg`. 👍

README.md
diegopetrucci: fixed https://github.com/Babylonpartners/Bento/pull/24/commits/32b15609a9a55a842e79cc722ce389a0284653f7

PR #30, opened 2018-05-23
Avoid name collisions while computing `reuseIdentifier`

dmcrodrigues:
### Context

`String(describing: View.self)` does not include any information about the namespace where it's defined leading to potential name collisions, which in this case means using the same identifier for views completely different.

Take as an example the following snippet where we define two components, `Text` and `Image`, and both define their inner view, `View` in their own namespace. The reuse identifier will be `"View"` i n both components. There's also the case where name collisions can appear between components defined in different modules.

```swift
enum Component {}

extension Component {

    final class Text: Renderable {
        func render(in view: View) {}
    }
}

extension Component.Text {

    public final class View: UIView {}
}

extension Component {

    final class Image: Renderable {

        func render(in view: View) {}
    }
}

extension Component.Image {

    public final class View: UIView {}
}
```

### Solution

`_typeName(_:qualified:)` is a Swift function that returns the demangled qualified name of a metatype which means will include the entire namespace where it's defined plus the module name which its part of.

This function is used by `String(reflecting:)` when the subject does not conform to neither of the following protocols:

- `CustomDebugStringConvertible`
- `CustomStringConvertible`
- `TextOutputStreamable`

Considering the example above this will lead to the following identifiers:

- `Component.Text`: `"{Module}.Component.Text.View"`
- `Component.Image`: `"{Module}.Component.Image.View"`

This avoids any collisions while reusing views associated to any of this components even when the actual view shares the same name.


andersio:
Shouldn't we use `String.init(reflecting:)` in this case, since this is the cannon public API?

PR #31, opened 2018-05-23
Allow custom table view adapter type.

andersio:
Introduce `UITableView.prepareForBoxRendering`.

This method prepares the UITableView for Bento `Box` rendering by:

1. initialising the adapter;
2. installing the adapter as the data source and the delegate; and
3. forcing the UITableView to reload immediately from the then-empty data source.

`prepareForBoxRendering` accepts either:

* metatypes representing the section and row IDs to be used; or
* the metatype of the custom adapter.
   [An example of a custom table view adapter.](https://github.com/Babylonpartners/Bento/pull/31/files#diff-798b0a8130d83b142dd89d9f2434e40bR7)

To maintain backward compatibility, it is not required to call `prepareForBoxRendering` prior to `render(_:)`. But `prepareForBoxRendering` would trap if any rendering has already happened.

`render(_:)` has also been made to trap if the ID types do not match.

---

### Implementation Note

`TableViewAdapterBase<U,  T>` not conforming `UITableViewDataSource` and `UITableViewDelegate` is a deliberate design decision.

ObjC-name-aware requirement matching work only at the conformance site. For example, if the base class conforms to the delegate protocols, and a subclass wants to implement `willDisplayCell:forRowAtIndexPath:`, the full name must be spelt using `@objc` i.e. `@objc(tableView:willDisplayCell:forRowAtIndexPath:)`. Otherwise, the method would not be exported with the expected Objective-C name, and therefore not be picked up by `UITableView`. The compiler also doesn't warn about the name mismatch in this scenario. So apparently in terms of API ergonomics, requiring the conformance on the subclass is a saner choice.


mluisbrown:
Looks good. Although it does pretty much tie our implementation to `UITableView` and exposes that via the API. Something we wanted to avoid 😢 But I can't see a better way right now.

It means that if we wanted to change the underlying implementation to use `UICollectionView`, for example, we could no longer do that without changing the public API.

andersio:
The Bento API right now is as an extension to `UITableView`, so it is already tightly coupled. 🤔

Bento/Bento/UITableViewExtensions.swift
```diff
@@ -0,0 +1,60 @@
+extension UITableView {
+    public func prepareForBoxRendering<SectionId: Hashable, RowId: Hashable>(
+        sectionIdType: SectionId.Type,
+        rowIdType: RowId.Type
+    ) {
+        prepareForBoxRendering(BentoTableViewAdapter<SectionId, RowId>.self)
+    }
+
+    public func prepareForBoxRendering<SectionId, RowId>(
+        _ type: TableViewAdapter<SectionId, RowId>.Type
```
sergdort: Would it be nicer to path instance here ? 🤔

andersio: Hmm.

andersio: This does. You specify the type of the adapter, and Bento would create it internally.

andersio: Both https://github.com/Babylonpartners/babylon-ios/pull/5317/ and 
[one of the examples](https://github.com/Babylonpartners/Bento/blob/56c4c6f310cbd227f9ea214f032cecb94ba871e5/Example/BookAppointmentViewController.swift) are using it.

sergdort: I believe it won't work if client has custom initializer

PR #32, opened 2018-05-23
Replace `_typeName(_:qualified:)` by `String(reflecting:)` which is safe for metatypes

dmcrodrigues:
`String(reflecting:)` is unsafe for instances since they can implement `CustomDebugStringConvertible` or `CustomStringConvertible` which could lead to dynamic values depending on the internal state but when applied to metatypes it's safe and can be used as the implementation for the default `reuseIdentifier`.


dmcrodrigues:
I guess it's fine 😛

PR #37, opened 2018-07-31
Make background of container views to have clear color

sergdort:
### Why

- Apparently default background color for `UITableViewCell` is `white`
- Sometimes we may want to have a components to have `clear` color (to have background color of `UITableView` or `UICollectionView`) like with our `InfoRenderer`


mluisbrown:
This might break existing users who depend on the white background color, no?

sergdort:
Yes, but I feel like this is a right way to do. Because for example user might want to set a custom view for a `UITableView`'s background view with a gradient for example

PR #38, opened 2018-08-02
Focus Control.

andersio:
This PR provides an infrastructure of focus coordination among Bento components.

### Creating a focusable component
For a Bento component to participate in the focus coordination, it would need to:

1. Conform to `Focusable`.

    Components may specify whether they participate in the focus coordination via `canBeFocused`. This is important for the focus coordination to discover all focusable components, especially those that are off screen.

2. Have its view conforming to `FocusableView`.

    The focus coordination invokes `focus()` on the conforming component root view when appropriate. The root view should respond [by asking the appropriate subviews (or itself) to assume the first responder status](https://github.com/Babylonpartners/Bento/blob/f00daad852eb02af7ecaa4a9d73815929da1e0f0/Example/Cells/TextFieldComponent.swift#L63).

    Conforming root views would automatically gain `withFocusCoordinator(_:)`, which provides scoped access to the focus coordination mechanism. The coordinator can be used to:

    a. [attempt to pass on its first responder status via `move(_:)`](https://github.com/Babylonpartners/Bento/blob/f00daad852eb02af7ecaa4a9d73815929da1e0f0/Example/Cells/TextFieldComponent.swift#L80); or

    b. [query whether there is any focusable component following it via `canMove(_:)`, so as to adjust the keyboard return key type, for example](https://github.com/Babylonpartners/Bento/blob/f00daad852eb02af7ecaa4a9d73815929da1e0f0/Example/Cells/TextFieldComponent.swift#L69).

3. Optionally be notified when a new 🍱 has been rendered.

    For example, if you have a input accessory toolbar with buttons to navigate between form fields, the availability of the buttons might be invalidated by new 🍱 boxes & therefore need to be updated. One may implement `neighbouringFocusEligibilityDidChange()` to receive such messages.

### Focus without user interaction

Bento introduces `focus(animated:)` to `UITableView` and `UICollectionView`, which upon invocation would scroll to the first focusable component & ask it to focus.

### Example: A personal details form.
Available as the fifth tab in the Bento example app. Check `PersonalDetailsViewController.swift` for the source.

### Implementation Note

This is architecturally the same as the focus control in our pre-Bento infrastructure. The notable difference is the implementation being extended to support both kind of collection views.

`FocusCoordinating` is a transient view of the abstract focus coordination mechanism.

It is exposed as a scoped accessor `withFocusCoordinator(_:)` to discourage escaping/retaining the coordinator. Prior Art: `withoutActuallyEscaping(_:)`.

Bento component views have zero knowledge about its position within the collection view. So for the purpose of focus coordination we need to query it & have it stored temporarily, but the info should be discarded as soon as we are done with the coordination.

The implementations introduce a group of internal protocols, centred by `BentoCollectionView`. `BentoContainerView` defines a set of requirements that is necessary for focus coordination (implemented as an extension to `BentoCollectionView`). Auxiliary protocols are introduced to either provide stronger constraints for container-agnostic implementations, or have a subset that can be used as existentials (protocol types) in dynamic type casting.

### DIFF
--- a/Bento/Adapters/BentoCollectionView.swift
+++ b/Bento/Adapters/BentoCollectionView.swift
@@ -1,0 +1,23 @@
+internal protocol BentoContainerCell {
+    var containedView: UIView? { get }
+}
+
+internal protocol BentoCollectionView: FocusCoordinatorProviding {
+    associatedtype Cell: BentoContainerCell
+    associatedtype DataSource
+
+    var dataSource: DataSource? { get }
+
+    func indexPath(for cell: Cell) -> IndexPath?
+    func visibleCell(at indexPath: IndexPath) -> Cell?
+    func revealCell(at indexPath: IndexPath, animated: Bool)
+}
+
+extension BentoCollectionView {
+    func didRenderBox() {
+        UIApplication.shared.sendAction(#selector(FocusableView.neighboringFocusEligibilityDidChange),
+                                        to: nil,
+                                        from: self,
+                                        for: nil)
+    }
+}
--- a/Bento/Helpers.swift
+++ b/Bento/Helpers.swift
@@ -49,3 +49,11 @@ extension Optional {
         return selector(unwrapped, other)
     }
 }
+
+internal func search<T>(from leaf: UIView, type: T.Type) -> T? {
+    var cell = leaf.superview
+    while let view = cell, !(view is T) {
+        cell = view.superview
+    }
+    return cell as! T?
+}
--- a/Bento/Renderable/AnyRenderable.swift
+++ b/Bento/Renderable/AnyRenderable.swift
@@ -1,14 +1,6 @@
 import UIKit
 
-struct AnyRenderable: Renderable, Deletable {
-    var canBeDeleted: Bool {
-        return base.canBeDeleted
-    }
-
-    var deleteActionText: String {
-        return base.deleteActionText
-    }
-
+struct AnyRenderable: Renderable {
     var reuseIdentifier: String {
         return base.reuseIdentifier
     }
@@ -17,10 +9,6 @@ struct AnyRenderable: Renderable, Deletable {
 
     init<Base: Renderable>(_ base: Base) where Base.View: UIView {
         self.base = AnyRenderableBox(base)
-    }
-
-    init<Base: Deletable>(_ base: Base) where Base.View: UIView {
-        self.base = AnyDeletableBox(base)
     }
 
     func generate() -> UIView {
@@ -67,30 +55,8 @@ struct AnyRenderable: Renderable, Deletable {
         return view
     }
 
-    func delete() {
-        base.delete()
-    }
-
     static func ==(lhs: AnyRenderable, rhs: AnyRenderable) -> Bool {
         return lhs.base.equals(to: rhs.base)
-    }
-}
-
-private class AnyDeletableBox<Base: Deletable>: AnyRenderableBox<Base> where Base.View: UIView {
-    override var canBeDeleted: Bool {
-        return base.canBeDeleted
-    }
-
-    override var deleteActionText: String {
-        return base.deleteActionText
-    }
-
-    override init(_ base: Base) {
-        super.init(base)
-    }
-
-    override func delete() {
-        base.delete()
     }
 }
 
@@ -126,14 +92,6 @@ private class AnyRenderableBox<Base: Renderable>: AnyRenderableBoxBase where Bas
 }
 
 private class AnyRenderableBoxBase {
-    var canBeDeleted: Bool {
-        return false
-    }
-
-    var deleteActionText: String {
-        return ""
-    }
-
     var reuseIdentifier: String { fatalError() }
 
     init() {}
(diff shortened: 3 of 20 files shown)


Bento/Helpers.swift
```diff
@@ -63,3 +63,11 @@ extension Optional {
         return selector(unwrapped, other)
     }
 }
+
+internal func search<T>(from leaf: UIView, type: T.Type) -> T? {
+    var cell = leaf.superview
+    while let view = cell, !(view is T) {
+        cell = view.superview
+    }
+    return cell as! T?
```
sergdort: Why not `cell as? T`?

andersio: This is an explicit assertion. It must be `T` or `nil` at this point.

Bento/Adapters/BentoCollectionView.swift
```diff
@@ -0,0 +1,23 @@
...
+    var dataSource: DataSource? { get }
+
+    func indexPath(for cell: Cell) -> IndexPath?
+    func visibleCell(at indexPath: IndexPath) -> Cell?
+    func revealCell(at indexPath: IndexPath, animated: Bool)
+}
+
+extension BentoCollectionView {
+    func didRenderBox() {
+        UIApplication.shared.sendAction(#selector(FocusableView.neighbouringFocusEligibilityDidChange),
```
mluisbrown: There's something so cool about being able to fire a message all the way up the responder chain 🚀

Bento/Renderable/AnyRenderable.swift
```diff
@@ -1,14 +1,10 @@
...
-struct AnyRenderable: Renderable, Deletable {
-    var canBeDeleted: Bool {
-        return base.canBeDeleted
-    }
+protocol ExistentialWrapper {
 
-    var deleteActionText: String {
-        return base.deleteActionText
-    }
+}
```
mluisbrown: [nit] extra line above. Actually since this is a marker protocol, could just be `{ }`

Bento/Renderable/FocusableView.swift
```diff
@@ -0,0 +1,25 @@
+@objc(BentoFocusableView)
+public protocol FocusableView: AnyObject {
+    @objc(bento_focus)
+    func focus()
+
+    @objc(bento_neighbouringFocusEligibilityDidChange)
+    optional func neighbouringFocusEligibilityDidChange()
```
mluisbrown: This needs to be documented. I don't fully understand it's use :/

[nit] possible should use US spelling (`neighboring`) to match Apple frameworks... 😷

andersio: Done.

Bento/Renderable/Deletable.swift
```diff
@@ -15,3 +13,15 @@ public extension Deletable {
         return nil
     }
 }
+
+// NOTE: Workaround to the lack of generalized existentials for Protocols with
+//       Associated Types (PATs).
+//
+//       See `_Focusable` for the rationale.
+/// Bento internal protocol. See `Deletable` for the public API.
+public protocol _Deletable {
```
sergdort: Why is this changed? 
It was working fine before no?

Bento/Adapters/SectionedFormAdapter.swift
```diff
@@ -97,11 +99,12 @@ final class SectionedFormAdapter<SectionId: Hashable, RowId: Hashable>
     @available(iOS 11.0, *)
     func tableView(_ tableView: UITableView, trailingSwipeActionsConfigurationForRowAt indexPath: IndexPath) -> UISwipeActionsConfiguration? {
         let row = sections[indexPath.section].rows[indexPath.row]
-        guard row.component.canBeDeleted else {
+        guard let component = row.component.cast(to: _Deletable.self),
```
sergdort: What was wrong with default implementation?

Bento/Renderable/Deletable.swift
andersio: I want to keep `Focusable` inheriting from `Renderable` as `Deletable` does. However, with the existing configuration it would require two extra existential wrappers (`AnyFocusableBox` and `AnyFocusableDeletableBox`), and even more stuff being dumped into `AnyRenderable`. This isn't a scalable solution if we are going to introduce more protocols, where the # of existential wrappers grows exponentially.

sergdort: Why do we need it to `inherit` from `Renderable`?

andersio: It isn't necessary for `Deletable`. But `Focusable` enforces a `View: FocusableView` constraint on the component.

andersio: Alternatively, we can drop both `Focusable where Self.View: FocusableView` and `Deletable: Renderable` constraints. Then these protocols would have no associated type and therefore can be used as existentials.

sergdort: Sounds good 👍 
We can also get rid of `Deletable` inherit from `Renderable`

sergdort:
Also I think we need to add documentation at least to read me that we have this functionality

PR #39, opened 2018-08-08
Support per component height estimation & height calculation.

andersio:
Allow components to supply estimated heights and actual heights, overriding the defaults specified by the `UITableView`.

1. Introduce two new optional requirements: `height(forWidth:)` and `estimatedHeight(forWidth:)`.

2. [Gardening] All `sizeBoundTo` methods have now been moved as an extension of `Renderable`, instead of being hosted by the `AnyRenderable` existential wrapper.

3. `sizeBoundTo(width:)` and `sizeBoundTo(size:)` would query the new optional requirements, and fallback to Auto Layout if necessary.

4. All sizing methods now accept optional inherited layout margins.

### DIFF
--- a/Bento/Adapters/SectionedFormAdapter.swift
+++ b/Bento/Adapters/SectionedFormAdapter.swift
@@ -54,6 +54,8 @@ final class SectionedFormAdapter<SectionId: Hashable, RowId: Hashable>
             componentView = component.generate()
             cell.install(view: componentView)
         }
+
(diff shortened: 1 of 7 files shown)


sergdort:
No cacheing?

sergdort:
My concern is that we keep adding stuff to `Renderable`, where `heigh` and `estimatedHeight` management could be done by client using custom adapter.

So I would rather merge #31 pr ¯\_(ツ)_/¯

PR #40, opened 2018-08-09
Extend the sizing API to support inherited layout margins.

andersio:
Cherry picked from #39.

This is important for computed sizes to be correct if the layout would be affected by layout margins inherited from superviews.


sergdort:
tenor-93634896

PR #42, opened 2018-08-12
Support collection supplementary views.

andersio:
### Section Supplement API
Sections now have a generic supplement API to support components attached to individual sections.

It differentiates supplements via the new `Supplement` enum key, which corresponds to header, footer or a custom string key (for custom UICollectionView layouts). In addition to the convenience initialisers, one may now also add or remove supplements via `Section.adding(_:_:)` and `Section.removing(_:_:)`.

If you render a Box carrying unsupported supplement, Bento data sources would gracefully ignore them.

This new set of API replaces all header- and footer-specific APIs exposed by `Section`, except for the convenience initialisers.

### UICollectionView Supplementary View support
With the new Section Supplement API, Bento now fully supports UICollectionView supplementary views with `UICollectionViewFlowLayout` and custom layouts.

### Bugfix: Crash due to component mismatch in UITableView Header/Footer.
The bugfix introduces a new shared code path for all Bento container views to handle component-view binding process, which now caters for the said type mismatch.

In other words, one may now use varying component types with the same item ID or the same supplement type without crashing.

Currently in production, Bento doesn't check the view type of the cell contained view before invoking `render(in:)` on the component.

Since UITableView does not re-request a new `UITableViewHeaderFooterView` for section header/footer views that are already on screen, this leads to a type mismatch & therefore crash when the component type for the given supplement type varies over time.

The 1st example tab has been updated to illustrate the bug.


Bento/Bento/Box.swift
```diff
@@ -22,30 +22,32 @@ infix operator |---*: NodeConcatenationPrecedence
 infix operator |---?: NodeConcatenationPrecedence
 infix operator <>: ComposingPrecedence
 
-public struct Box<SectionId: Hashable, RowId: Hashable> {
-    public let sections: [Section<SectionId, RowId>]
+public struct Box<SectionID: Hashable, ItemID: Hashable> {
+    public typealias Section = Bento.Section<SectionID, ItemID>
```
mluisbrown: ❤️

sergdort:
@andersio Shall we also make `CollectionViewDataSource` open?

andersio:
@sergdort That shall be a separate task.

Bento/Adapters/CollectionViewDataSource.swift
```diff
@@ -43,26 +46,43 @@ final class CollectionViewDataSource<SectionId: Hashable, ItemId: Hashable>
...
+            collectionView.register(CollectionViewContainerReusableView.self,
+                                    forSupplementaryViewOfKind: kind,
+                                    withReuseIdentifier: reuseIdentifier)
+            let view = collectionView.dequeueReusableSupplementaryView(
+                ofKind: kind,
+                withReuseIdentifier: reuseIdentifier,
+                for: indexPath
+            ) as! CollectionViewContainerReusableView
+
+            view.bind(component)
```
sergdort: 👍

PR #44, opened 2018-08-21
Grouped table view header/footer color

sergdort:
As per UIKit logs:

`Setting the background color on UITableViewHeaderFooterView has been deprecated. Please set a custom UIView with your desired background color to the backgroundView property instead.`

We need to set custom view to the background view of `UITableViewHeaderFooterView` to achieve custom color.


Bento/Views/TableViewHeaderFooterView.swift
```diff
@@ -9,6 +9,9 @@ final class TableViewHeaderFooterView: UITableViewHeaderFooterView {
 
     override init(reuseIdentifier: String?) {
         super.init(reuseIdentifier: reuseIdentifier)
+        let view = UIView()
```
mluisbrown: I guess `UITableView` takes care of layout etc?

PR #45, opened 2018-08-21
Use contentView when appropriate. Fixed supplementary view updates.

andersio:
1. Container views should use `contentView` whenever available so as to be able to participate in self sizing by default.

2. Handle missing supplement correctly.

    UICollectionView accepts only pre-registered types being returned for supplementary views. So now if Bento does not manage to find the requested supplement from the given section, it would dequeue an empty reusable view using `emptyReuseIdentifier`.

3. Handle supplementary view update correctly.

    Since the item index of the index paths to supplementary view is layout specific, `CollectionViewSectionDiff` now first queries `indexPathsForVisibleSupplementaryElements` on `UICollectionView` to gather all index paths. It would then proceed to obtain the view and perform necessary updates.

4. Make sure UITableView/UICollectionView supplementary views are updated for moved sections.

    The update logic currently considers only in-place section mutation.


Bento/Adapters/CollectionViewDataSource.swift
```diff
@@ -61,25 +63,23 @@ final class CollectionViewDataSource<SectionID: Hashable, ItemID: Hashable>
...
-
-            view.bind(component)
-            return view
-        } else {
-            return UICollectionReusableView()
-        }
+        let component = sections[indexPath.section].supplements[supplement]
+        let reuseIdentifier = component?.reuseIdentifier ?? emptyReuseIdentifier
+
+        collectionView.register(CollectionViewContainerReusableView.self,
```
sergdort: So here we are registering view every time, not sure if it has some impact on performance,

@dmcrodrigues sugested to track registered views (may be in a dictionary or something)

Bento/Diff/Changeset+MutationIndexPairs.swift
```diff
@@ -0,0 +1,10 @@
+import FlexibleDiff
+
+extension Changeset {
+    var mutationIndexPairs: FlattenCollection<[[(source: Int, destination: Int)]]> {
```
sergdort: 👍

Bento/Adapters/CollectionViewDataSource.swift
mluisbrown: AFAIK the internal logic for `register(_: forSupplementaryViewOfKind:withReuseIdentifier:)` uses a dictionary, so it wouldn't make much difference.

PR #46, opened 2018-09-06
Reload collection view if it's not in the window hierarchy

sergdort:
### Description

Apparently collection view does not like batch updates if it's not in the window hierarchy ¯\_(ツ)_/¯ 

### How

- Reload collection view if it's not in the window hierarchy


Bento/Adapters/CollectionViewDataSource.swift
```diff
@@ -35,6 +31,12 @@ final class CollectionViewDataSource<SectionID: Hashable, ItemID: Hashable>
 
     func update(sections: [Section<SectionID, ItemID>], completion: (() -> Void)?) {
         guard let collectionView = collectionView else { return }
+        if collectionView.window == nil { // Just reload collection view if it's not in the window hierarchy
```
andersio: [Nit] Comment on next line please?

PR #47, opened 2018-09-07
Remove return from the loop

sergdort:
Remove a return from the loop. As we may need to bind cell on next iteration


andersio:
Also in CollectionViewSectionDiff. 🤦‍♂️

PR #56, opened 2018-10-24
Introduce BentoKit

dmcrodrigues:


.gitattributes
```diff
@@ -0,0 +1,183 @@
+Snapshots/ReferenceImages_64/BentoKitTests.ToggleSnapshotTests/test_toggle_loading_iPhone5@2x.png filter=lfs diff=lfs merge=lfs -text
```
andersio: Shouldn't we use a wildcard here?

dmcrodrigues:
@sergdort please review this.

PR #57, opened 2018-10-25
Switch to pre-built binaries

dmcrodrigues:
Xcode 10 does not support well `SUPPORTED_PLATFORMS` leading to random failures while building.

This should fix `master` which is currently failing because CI it's trying to build for macOS.


dmcrodrigues:
We decide to not pursue this path so I'm closing it.

PR #58, opened 2018-10-25
Add ComponentLifecycle and expose AnyRenderable

sergdort:
## Why

- we may need to know when component appears and disappears from the screen (e.g analytics)
- we may want to know when a root view of the component will be presented (e.g to be able to start animation)

## How

- Add `ComponentLifecycleAware` protocol which client can conform to be notified when component gonna be shown

```swift
public protocol ComponentLifecycleAware {
    func willDisplayItem()
    func didEndDisplayingItem()
}
```
- Add `ViewLifecycleAware` protocol which client can conform to be notified when root view of the component gonna be shown/hidden
```swift
public protocol ViewLifecycleAware {
    func willDisplayView()
    func didEndDisplayingView()
}

- Expose `AnyRenderable` so users of the library will be able to implement advanced behaviour similar to this
```


Bento/Views/BentoReusableView.swift
```diff
@@ -20,6 +22,20 @@ extension BentoReusableView {
             containedView = nil
         }
     }
+
+    func willDisplayView() {
```
dmcrodrigues: As discussed I think we can move this methods to be part of `TableViewContainerCell` and `CollectionViewContainerCell` which hold both, `component` and `containedView`, and can invoke/forward this methods when they need to be aware of the lifecycle and that way we avoid exposing `AnyRenderable`.

Bento/Views/BentoReusableView.swift
```diff
@@ -50,3 +71,15 @@ extension BentoReusableView where Self: UIView {
         }
     }
 }
+
+private class Value<T>: NSObject {
```
dmcrodrigues: I believe this is no longer needed.

Bento/Views/BentoReusableView.swift
```diff
@@ -20,6 +27,20 @@ extension BentoReusableView {
             containedView = nil
```
dmcrodrigues: Maybe we should release the component here? In terms of correctness if there's no `containedView` we shouldn't have a `component` as well.

sergdort: well, the incoming component is optional already so if it's `nil` the it will be set to `nil` 😛

Example/BookAppointmentViewController.swift
```diff
@@ -5,7 +5,7 @@ import ReactiveFeedback
 import enum Result.NoError
 
 final class CustomTableViewAdapter<SectionId: Hashable, RowId: Hashable>: TableViewAdapterBase<SectionId, RowId>, UITableViewDataSource, UITableViewDelegate {
-    func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
+    override func tableView(_ tableView: UITableView, willDisplay cell: UITableViewCell, forRowAt indexPath: IndexPath) {
```
andersio: Please call `super`.

PR #60, opened 2018-11-01
Add support for `UIMenuController` actions when user long press item in [UI]TableView/CollectionView 

sergdort:
## Description 

We need to have ability to add custom actions to `UIMenuController`

#### How UIKit manages `UIMenuController` with custom actions:

1. Users need to set array of actions they want to add

```swift
UIMenuController.shared.menuItems = [UIMenuItem(title: "Report", action: #selector(someMethod))]
```

2. Then [UI]CollectionView/TableView delegate should implement all 3 methods related to menu items

```swift
func collectionView(_ collectionView: UICollectionView, shouldShowMenuForItemAt indexPath: IndexPath) -> Bool

func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath, withSender sender: Any?) -> Bool 

public func collectionView(_ collectionView: UICollectionView, performAction action: Selector, forItemAt indexPath: IndexPath, withSender sender: Any?)
```

3. Then UIKit will ask `UICollectionViewCell` whether it responds to that selector (if not UIKit will ignore and won't show custom action in the pop up)

## Proposed solution

1. Introduce `MenuItemsResponding`
```swift
public protocol MenuItemsResponding: NSObjectProtocol {
    var menuItems: [UIMenuItem] { get }
}
```

As you may notice we need to have `NSObjectProtocol` (we'll need it later to use for message forwarding)

2. Then in `CollectionViewContainerCell` we simply forward all messages that associated with `MenuItemsResponding` to the component object

```swift
final class CollectionViewContainerCell: UICollectionViewCell {
    override func responds(to aSelector: Selector!) -> Bool {
        guard let component = component?.cast(to: MenuItemsResponding.self) else {
            return super.responds(to: aSelector)
        }

        return component.responds(to: aSelector) || super.responds(to: aSelector)
    }

    override func forwardingTarget(for aSelector: Selector!) -> Any? {
        guard let component = component?.cast(to: MenuItemsResponding.self) else {
            return super.forwardingTarget(for: aSelector)
        }

        return component
    }
}
```

3. What we need from the client side:

- Define a component which conforms to `MenuItemsResponding ` and add a `@objc` method to it

```swift
extension Component {
    final class IncomingMessage: NSObject, AutoRenderable, MenuItemsResponding {
       var menuItems: [UIMenuItem] {
              return [
                     UIMenuItem(title: "Report", action: #selector(didReportItem))
              ]
        }

        @objc
        func didReportItem() {
            print(#function)
        }
    }
}
```

## How does it look:

custom_menu_item

### DIFF
--- a/Bento.xcodeproj/project.pbxproj
+++ b/Bento.xcodeproj/project.pbxproj
@@ -13,6 +13,7 @@
 		5829D26E200FB092001E020D /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 5829D26C200FB092001E020D /* Main.storyboard */; };
 		5829D270200FB092001E020D /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 5829D26F200FB092001E020D /* Assets.xcassets */; };
 		5829D273200FB092001E020D /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 5829D271200FB092001E020D /* LaunchScreen.storyboard */; };
+		582D9951217E196E00C67B0D /* MenuItemsResponding.swift in Sources */ = {isa = PBXBuildFile; fileRef = 582D9950217E196E00C67B0D /* MenuItemsResponding.swift */; };
 		582D9987217F87B100C67B0D /* ComponentLifecycleAware.swift in Sources */ = {isa = PBXBuildFile; fileRef = 582D9986217F87B100C67B0D /* ComponentLifecycleAware.swift */; };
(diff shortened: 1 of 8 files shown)


dmcrodrigues:
In your example in point 3) with `IncomingMessage`, where are you defining/installing the menu items in `UIMenuController.shared`? And how does this works if you have difference items/cells with different menu items?

sergdort:
@dmcrodrigues In 

`collectionView(_ collectionView: UICollectionView, shouldShowMenuForItemAt indexPath: IndexPath) -> Bool`

Collection view asks if we can perform an action for item at `indexPath` this happens during long press (as it about to show a menu item). So this is where we add custom items

@andersio If you do not want to have a custom adapter that we can wait for your PR to be merged and I move these methods into our adapter.

PR #63, opened 2018-11-02
Support for `NavigationTitleItem` in `Screen`. Resolves #61

### LINKED ISSUE #61: `Screen` and `BoxViewController` don't support an image for the navigation title
Currently `Screen` and, by extension, `BoxViewController` only support text for the navigation bar title. They should support images as well.

`Screen` already has `NavigationTitleItem` for this, but it is not yet being used.

There should be a convenience initialiser for `Screen` which takes `NavigationTitleItem` instead of a `String` for the title, and `BoxViewController` needs to handle setting the image, in case the title is set as an image.

mluisbrown:
Adds support for using `NavigationTitleItem` in `Screen`, so that you can use images (or any other `UIView` for the navigation title.

Snapshot tests updated.

Resolves #61

### DIFF
--- a/BentoKit/BentoKit/Screen/Screen.swift
+++ b/BentoKit/BentoKit/Screen/Screen.swift
@@ -37,7 +37,7 @@ public struct Screen<SectionId: Hashable, RowId: Hashable> {
     }
 
     public init(
-        titleItem: NavigationTitleItem = .text(""),
+        titleItem: NavigationTitleItem,
         leftBarItems: [BarButtonItem] = [],
         rightBarItems: [BarButtonItem] = [],
         formStyle: BentoTableView.Layout = .topYAligned,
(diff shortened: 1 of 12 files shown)


BentoKit/BentoKit/Screen/Screen.swift
```diff
@@ -22,10 +22,34 @@ public struct Screen<SectionId: Hashable, RowId: Hashable> {
...
+            focusMode: focusMode,
+            shouldUseSystemSeparators: shouldUseSystemSeparators,
+            box: box,
+            pinnedToTopBox: pinnedToTopBox,
+            pinnedToBottomBox: pinnedToBottomBox
+        )
+    }
+
+    public init(
+        titleItem: NavigationTitleItem = .text(""),
```
dmcrodrigues: Can we please change this to make it explicit for consumers they need to pass a `title` or `titleItem`? Having an empty one should be defined by them not us.

mluisbrown: I can make `titleItem` have no default value, but `title` (in the old initialiser) always had a default value. Removing it now would break a lot of existing code.

dmcrodrigues: I still think it should be done, this needs to be defined explicitly and not implicitly.

mluisbrown: Done 💥

dmcrodrigues: ❤️

PR #67, opened 2018-11-06
Add ComponentLifecycle aware conformance for TitledDescription

sergdort:
Because of the long transition `TitledDescription` in our main repo ended up having `ComponentLyfecycleAware` behaviour while the one here does not ¯\_(ツ)_/¯


mluisbrown:
You need to actually add the methods of the protocol too 😄 

See `willDisplayItem` and `didEndDisplayingItem` in the non OSS version.

sergdort:
🤦‍♂️

PR #68, opened 2018-11-06
Include the asset catalog in the pod resource bundle

mluisbrown:
Includes the asset catalog in the pod resource bundle. 

Change the code to use the pod resource bundle.

### DIFF
--- a/BentoKit.podspec
+++ b/BentoKit.podspec
@@ -16,7 +16,7 @@ Pod::Spec.new do |s|
 # Switch back to use the version tag when we release 0.4.
   s.source           = { :git => "https://github.com/Babylonpartners/Bento.git", :branch => "master" } #, :tag => "#{s.version}" }
   s.source_files     = 'BentoKit/BentoKit/**/*.swift'
-  s.resources        = ['BentoKit/BentoKit/Resources/Media.xcassets']
+  s.resource_bundles =  { 'BentoKit' => 'BentoKit/BentoKit/Resources/Media.xcassets' }
   s.exclude_files    = 'BentoKit/BentoKit/Components.playground/**/*'
 
   s.swift_version = "4.2"
--- a/BentoKit.podspec
+++ b/BentoKit.podspec
@@ -14,9 +14,9 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '10.0'
 
 # Switch back to use the version tag when we release 0.4.
-  s.source           = { :git => "https://github.com/Babylonpartners/Bento.git", :branch => "develop" } #, :tag => "#{s.version}" }
+  s.source           = { :git => "https://github.com/Babylonpartners/Bento.git", :branch => "master" } #, :tag => "#{s.version}" }
   s.source_files     = 'BentoKit/BentoKit/**/*.swift'
-  s.resource_bundles =  { 'BentoKit' => 'BentoKit/BentoKit/Resources/Media.xcassets' }
+  s.resources        = ['BentoKit/BentoKit/Resources/Media.xcassets']
   s.exclude_files    = 'BentoKit/BentoKit/Components.playground/**/*'
 
   s.swift_version = "4.2"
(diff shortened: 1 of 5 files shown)


BentoKit.podspec
```diff
@@ -14,9 +14,10 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '10.0'
 
 # Switch back to use the version tag when we release 0.4.
-  s.source        = { :git => "https://github.com/Babylonpartners/Bento.git", :branch => "develop" } #, :tag => "#{s.version}" }
-  s.source_files  = 'BentoKit/BentoKit/**/*.swift'
-  s.exclude_files = 'BentoKit/BentoKit/Components.playground/**/*'
+  s.source           = { :git => "https://github.com/Babylonpartners/Bento.git", :branch => "develop" } #, :tag => "#{s.version}" }
```
sergdort: I do not think we have `develop` branch

BentoKit/BentoKit/Views/FocusToolbar.swift
```diff
@@ -8,7 +8,9 @@ public final class FocusToolbar: UIToolbar {
     private var view: (UIView & FocusableView)?
 
     public init(view: UIView & FocusableView) {
-        let bundle = Bundle(for: FocusToolbar.self)
+        let url = Bundle(for: FocusToolbar.self).url(forResource: "BentoKit", withExtension: "bundle")
```
sergdort: So cocoa pods do not just add asset catalog to the framework target?

sergdort: I mean, why can't we just use `Bundle(for: FocusToolbar.self)`?

BentoKit.podspec
mluisbrown: This was already there, but I'll change it to `master`.

BentoKit/BentoKit/Views/FocusToolbar.swift
mluisbrown: If you use `resource_bundles` in the podspec it puts the resources into a separate bundle when making the pod version. I'll try with just `resources` to see if this avoids having one path for running locally and one for when part of the pod.

andersio:
Static libraries are considered part of the application bundle by nature, which if BentoKit is packaged as su means these assets would be copied by CocoaPods into the application bundle, causing potential clashes with clients.

mluisbrown:
@andersio I previously had it configured using `resource_bundles`, which means you have to load the resources differently depending on whether BentoKit was built as part of a pod or not. See 
https://github.com/Babylonpartners/Bento/pull/68/commits/3d6cda39cb06eee8c702ffd1b266b41bc7cb0946 .

Is it likely that BentoKit will ever be linked as a static library? IMO it isn't 🤷🏼‍♀️

andersio:
`use_frameworks!` is no longer mandatory with CocoaPods supporting Swift static library. While I am not sure exactly if the behaviour differs if a pod specifies resources, AFAIK it would not build pods as frameworks in its absence at least based on my hands-on experience with source only pods e.g. RAS. There is a strong case not only for other users, but for ourselves to move to static linkage to lower cold launch time.

mluisbrown:
@andersio Ok, I've reverted to using `resource_bundles` in the podspec 👍

BentoKit/BentoKit/Views/Resources.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+final class Resources {
```
mluisbrown: This has to be a class so that it can be passed to `Bundle(for:)`

sergdort:
@andersio I thought that you can not have static library if you have resources

andersio:
@sergdort You can. CocoaPods would just copy the resources to the application bundle, and it already does the same thing pre-iOS 8 when there is no dylib.

PR #69, opened 2018-11-07
Added search bar component

ilyapuchka:
This PR adds a simple component that contains `UISearchBar`. 

### Motivation

Sometimes it might be needed just to display a search bar in the cell without the luggage of `UISearchViewController` which might be hard sometimes to tweak to fit particular UX, and something that requires less customisation than `UITextField` to make it look like `UISearchBar`.

### Alternatives considered

There is currently `TextInput` component in BentoKit but it kind of serves a different purpose and also requires quite some tweaking and style additions to make it fit this use case. Also `UISearchBar` is not a subclass of `UITextField` and have some extra features like showing cancel button and search icon out of the box (still could be done with `TextInput` but not without more changes to it).

`UISearchViewController` also does not fit here very well as it requires `navigationItem` and it can be tricky to make it work with DrawerKit used here (see screenshot)

### Example

simulator screen shot - iphone xs max - 2018-11-06 at 19 40 20

### DIFF
--- a/BentoKit/BentoKit/Components/Search.swift
+++ b/BentoKit/BentoKit/Components/Search.swift
@@ -12,7 +12,7 @@ extension Component {
             didBeginEditing: Optional<(UISearchBar) -> Void> = nil,
             textDidChange: Optional<(UISearchBar, String) -> Void> = nil,
             cancelButtonClicked: Optional<(UISearchBar) -> Void> = nil,
-            styleSheet: StyleSheet
+            styleSheet: StyleSheet = StyleSheet()
             ) {
             self.configurator = { view in
                 view.searchBar.placeholder = placeholder
--- a/BentoKit/BentoKit/Components/Search.swift
+++ b/BentoKit/BentoKit/Components/Search.swift
@@ -9,13 +9,15 @@ extension Component {
         public init(
             placeholder: String? = nil,
             keyboardType: UIKeyboardType = .default,
-            textDidChange: Optional<(String) -> Void> = nil,
-            cancelButtonClicked: Optional<() -> Void> = nil,
+            didBeginEditing: Optional<(UISearchBar) -> Void> = nil,
+            textDidChange: Optional<(UISearchBar, String) -> Void> = nil,
+            cancelButtonClicked: Optional<(UISearchBar) -> Void> = nil,
             styleSheet: StyleSheet
             ) {
             self.configurator = { view in
                 view.searchBar.placeholder = placeholder
                 view.searchBar.keyboardType = keyboardType
+                view.didBeginEditing = didBeginEditing
                 view.textDidChange = textDidChange
                 view.cancelButtonClicked = cancelButtonClicked
             }
@@ -29,9 +31,9 @@ extension Component.Search {
 
         fileprivate let searchBar = UISearchBar()
 
-        var textDidChange: Optional<(String) -> Void> = nil
-        var cancelButtonClicked: Optional<() -> Void> = nil
-        var didBeginEditing: Optional<() -> Void> = nil
+        var textDidChange: Optional<(UISearchBar, String) -> Void> = nil
+        var cancelButtonClicked: Optional<(UISearchBar) -> Void> = nil
+        var didBeginEditing: Optional<(UISearchBar) -> Void> = nil
 
         public override init(frame: CGRect) {
             super.init(frame: frame)
@@ -51,7 +53,7 @@ extension Component.Search {
         }
 
         public func searchBarTextDidBeginEditing(_ searchBar: UISearchBar) {
-            self.didBeginEditing?()
+            self.didBeginEditing?(searchBar)
         }
 
         public func searchBarSearchButtonClicked(_ searchBar: UISearchBar) {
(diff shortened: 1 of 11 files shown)


BentoKit/BentoKit/Components/Search.swift
```diff
@@ -0,0 +1,126 @@
...
+        }
+    }
+}
+
+extension Component.Search {
+    public final class View: BaseView, UISearchBarDelegate {
+
+        let searchBar = UISearchBar()
+
+        var textDidChange: Optional<(UISearchBar, String) -> Void> = nil
```
sergdort: Do we really need the `UISearchBar` there?

BentoKit/BentoKit/Components/Search.swift
```diff
@@ -0,0 +1,126 @@
...
+
+extension Component {
+    public final class Search: AutoRenderable {
+        public let configurator: (View) -> Void
+        public let styleSheet: StyleSheet
+
+        public init(
+            placeholder: String? = nil,
+            keyboardType: UIKeyboardType = .default,
+            didBeginEditing: Optional<(UISearchBar) -> Void> = nil,
```
sergdort: dito

BentoKit/BentoKit/Components/Search.swift
```diff
@@ -0,0 +1,126 @@
...
+
+        public func searchBarCancelButtonClicked(_ searchBar: UISearchBar) {
+            self.cancelButtonClicked?(searchBar)
+        }
+    }
+}
+
+extension Component.Search {
+    public final class StyleSheet: BaseViewStyleSheet<View> {
+        public struct SearchBar: StyleSheetProtocol {
```
sergdort: This should be moved to the `StyleSheets` framework. And should be extended from `ViewStyleSheet` as `UISearchBar` is a subclass of `UIView`

BentoKit/BentoKit/Components/Search.swift
ilyapuchka: I decided to follow the signature of the delegate method. also it makes it easier to do things like dismiss keyboard when cancel is pressed (for some reason is not a default behaviour) without jumping through the hoops.

BentoKit/BentoKit/Components/Search.swift
ilyapuchka: @sergdort done 👍

ilyapuchka:
@sergdort I've addressed your comments

BentoKit/BentoKit/Components/Search.swift
sergdort: fair enough

StyleSheets/StyleSheets/SearchBarStyleSheet.swift
```diff
@@ -0,0 +1,66 @@
...
+                                            cornerRadius: cornerRadius)
+        element.showsCancelButton = showsCancelButton
+        element.searchTextPositionAdjustment = searchTextPositionAdjustment
+        element.keyboardType = keyboardType
+        element.returnKeyType = returnKeyType
+        element.enablesReturnKeyAutomatically = enablesReturnKeyAutomatically
+        element.barTintColor = tintColor
+        // remove system background
+        element.backgroundColor = .white
+        element.backgroundImage = UIImage()
```
sergdort: This should be a property of the stylesheet IMO as some may want to customise it as well

BentoKit/BentoKit/Components/Search.swift
```diff
@@ -0,0 +1,82 @@
...
+
+        public func searchBarCancelButtonClicked(_ searchBar: UISearchBar) {
+            self.cancelButtonClicked?(searchBar)
+        }
+    }
+}
+
+extension Component.Search {
+    public final class StyleSheet: BaseViewStyleSheet<View> {
+        public var searchBar = SearchBarStyleSheet()
```
sergdort: this can be let

StyleSheets/StyleSheets/SearchBarStyleSheet.swift
```diff
@@ -0,0 +1,66 @@
...
+        self.returnKeyType = returnKeyType
+        self.enablesReturnKeyAutomatically = enablesReturnKeyAutomatically
+        super.init(backgroundColor: backgroundColor, cornerRadius: cornerRadius)
+    }
+
+    open override func apply(to element: UISearchBar) {
+        super.apply(to: element)
+        element.setTextInputBackgroundColor(color: backgroundColor ?? .clear,
+                                            height: height,
+                                            cornerRadius: cornerRadius)
```
sergdort: Why do we need this

StyleSheets/StyleSheets/SearchBarStyleSheet.swift
ilyapuchka: I'm not sure, this is needed just to make these weird lines go away. Here how it looks without it.
screen shot 2018-11-08 at 15 38 14

The background colour can be configured though so I'll adjust style to configure it separately for text field.

sergdort: ok

StyleSheets/StyleSheets/SearchBarStyleSheet.swift
ilyapuchka: Because on `UISearchBar` one can only set the background image, not a colour.

ilyapuchka:
@sergdort addressed 😉

PR #70, opened 2018-11-07
Fix issue with bar button item not resigning first responder 

sergdort:
## Description

There is an issue with BoxView when sometimes it does not resign first responder when bar button pressed leaving any editing that was in progress uncommitted.

## Solution

Make sure we resign first responder before sending bar button actions

### DIFF
--- a/BentoKit/BentoKit/Screen/Screen.swift
+++ b/BentoKit/BentoKit/Screen/Screen.swift
@@ -132,9 +132,11 @@ public struct BarButtonItem {
         return item
     }
 
-    public func applyNonAppearanceChanges(to item: UIBarButtonItem) {
-        item.didTap = callback
-        item.isEnabled = isEnabled
+    public func applyNonAppearanceChanges(to item: UIBarButtonItem, willTriggerAction: (() -> Void)? = nil) {
+        item.didTap = {
+            willTriggerAction?()
+            self.callback?()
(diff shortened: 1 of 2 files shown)


BentoKit/BentoKit/Screen/Screen.swift
```diff
@@ -132,9 +132,11 @@ public struct BarButtonItem {
         return item
     }
 
-    public func applyNonAppearanceChanges(to item: UIBarButtonItem) {
-        item.didTap = callback
-        item.isEnabled = isEnabled
+    public func applyNonAppearanceChanges(to item: UIBarButtonItem, willTriggerAction: (() -> Void)? = nil) {
+        item.didTap = {
+            willTriggerAction?()
+            self.callback?()
```
mluisbrown: Where was the callback being called before?

BentoKit/BentoKit/Screen/Screen.swift
```diff
@@ -132,9 +132,11 @@ public struct BarButtonItem {
         return item
     }
 
-    public func applyNonAppearanceChanges(to item: UIBarButtonItem) {
-        item.didTap = callback
-        item.isEnabled = isEnabled
```
sergdort: OMG this should not be deleted 🤦‍♂️

PR #72, opened 2018-11-08
Option to disable first responder ascension upon interaction.

andersio:
In general, as a safe default, we would want all components to call back only after ascending themselves to be the first responder. This allows the first responder (text controls in particular) to first write back the data.

But there are scenarios where such behaviour is unnecessary and causing keyboard animation glitches. For example, when using Bento as an autocompletion list augmenting an input view, this would lead to the input view being undesirably dismissed upon user selecting a suggestion. One might instead want to keep the input view active, so as to allow users to continue typing.

This PR adds the ability to opt out in a variety of BentoKit components to disable such behaviour. It is represented with the new `InteractionBehavior` option set, where the opt-out can be specified as an (empty) set without the `becomeFirstResponder` option.

### DIFF
--- a/BentoKit/BentoKit/Components/TitledDescription.swift
+++ b/BentoKit/BentoKit/Components/TitledDescription.swift
@@ -43,7 +43,6 @@ public extension Component {
             image: Property<BentoKit.ImageOrLabel>? = nil,
             accessory: Accessory = .chevron,
             badgeIcon: UIImage? = nil,
-            isEnabled: Bool = true,
             inputNodes: CustomInput? = nil,
             didTap: Optional<() -> Void> = nil,
             didTapAccessory: Optional<() -> Void> = nil,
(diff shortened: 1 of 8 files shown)


sergdort:
Should we have for all other components?

BentoKit/BentoKit/Components/TitledDescription.swift
```diff
@@ -43,7 +43,6 @@ public extension Component {
             image: Property<BentoKit.ImageOrLabel>? = nil,
             accessory: Accessory = .chevron,
             badgeIcon: UIImage? = nil,
-            isEnabled: Bool = true,
```
Ben-Henshall: Why was isEnabled removed? Was it not being used by anything?

andersio: not used as flagged by @mluisbrown

PR #73, opened 2018-11-09
Show search bar cancel button with animation

ilyapuchka:
`SearchBarStyleSheet` applies cancel button without animation, so when screen is re-rendered with this property updated no animation happens in the search bar. This will perform animation by default. But we might want to come up with a more generic solution that could be used for all properties that can be set with animation, i.e. instead of using primitive types like `Bool` here we could introduce something like (not necessary an enum)

```swift
enum AnimatableProperty<T> {
  case withAnimation(T)
  case withoutAnimation(T)
}
```


ilyapuchka:
@andersio makes sense. I've done this change

PR #75, opened 2018-11-15
Allow providing custom views as accessory

ilyapuchka:
This will allow to provide a custom view, i.e. button, as accessory view. Here it's a button with "call" title and icon on the right side.

screen shot 2018-11-15 at 18 21 57

### DIFF
--- a/BentoKit/BentoKit/Views/Accessory.swift
+++ b/BentoKit/BentoKit/Views/Accessory.swift
@@ -42,6 +42,10 @@ public final class AccessoryView: InteractiveView {
     }
 
     public override var intrinsicContentSize: CGSize {
+        if case let .custom(view) = accessory {
+            view.layoutIfNeeded()
+            return view.frame.size
+        }
         return CGSize(width: 24, height: 24)
     }
 
@@ -63,6 +67,8 @@ public final class AccessoryView: InteractiveView {
         case let .tintedIcon(image, tintColor):
             view = UIImageView(image: image)
                 .with { $0.tintColor = tintColor }
+        case let .custom(customView):
+            view = customView
         case .none:
             view = nil
         }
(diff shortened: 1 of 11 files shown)


BentoKit/BentoKit/Views/Accessory.swift
```diff
@@ -42,6 +42,10 @@ public final class AccessoryView: InteractiveView {
     }
 
     public override var intrinsicContentSize: CGSize {
+        if case let .custom(view) = accessory {
+            view.layoutIfNeeded()
```
sergdort: I think if `view` would pinned to the edges not to `centerInSuperview` them you would not need that

ilyapuchka: what about width?
Actually I think there needs to be an adjustment for the height calculation as it currently has some constant `24`. Though it seems to be working in playground correctly.

andersio: Why isn't the intrinsic content size of the custom view be used? If one needs an accessory that cannot have a definitive intrinsic content size, it sounds something is off.

ilyapuchka:
@andersio When I tried to return just this view intrinsic content size does not work, i.e. if one tries to use a UIButton. Though this was my original idea. I'll add a screenshot.

ilyapuchka:
In fact this is what happening in test with `return view.intrinsicContentSize`

screen shot 2018-11-16 at 18 48 37

And here correct layout with `view.layoutIfNeeded`

screen shot 2018-11-16 at 18 48 15

BentoKit/BentoKit/Views/Accessory.swift
andersio: Fair enough.

PR #76, opened 2018-11-16
Refactor TitledDescription styling

sergdort:
## Why

I need to be able to make `TitledDescription` to look like this:

screenshot 2018-11-16 at 14 48 27

## How

- Added `content` stylesheet to the `TitledDescription.Stylesheet`
- Remove all related properties of the content and use `content` stylesheet
- Add `borderWidth` and `borderColor` properties to the `BaseStackView`

### DIFF
--- a/BentoKit/BentoKit/Components/TitledDescription.swift
+++ b/BentoKit/BentoKit/Components/TitledDescription.swift
@@ -538,7 +538,7 @@ public extension Component.TitledDescription {
         }
     }
 
-    public final class StyleSheet: BaseViewStyleSheet<View> {
+    public final class StyleSheet: InteractiveViewStyleSheet<View> {
 
         public enum HighlightingTarget {
             case container
--- a/BentoKit/BentoKit/Components/TitledDescription.swift
+++ b/BentoKit/BentoKit/Components/TitledDescription.swift
@@ -189,10 +189,10 @@ private extension Component.TitledDescription {
         return { width, inheritedMargins in
             guard width > 0 else { return 0 }
 
-            let xSpacing = styleSheet.horizontalSpacingBetweenElements
+            let xSpacing = styleSheet.content.spacing
             let verticalMargins = styleSheet.layoutMargins.verticalTotal
-                + (styleSheet.contentStackViewLayoutMarginsRelativeArrangement
-                    ? styleSheet.contentLayoutMargins.verticalTotal
+                + (styleSheet.content.isLayoutMarginsRelativeArrangement
+                    ? styleSheet.content.layoutMargins.verticalTotal
                     : 0)
 
             let detailWidth = detail?.width(using: styleSheet.detail) ?? 0
@@ -209,8 +209,8 @@ private extension Component.TitledDescription {
             let availableWidthForLabelBlock = width
                 - max(styleSheet.layoutMargins.left, inheritedMargins.left)
                 - max(styleSheet.layoutMargins.right, inheritedMargins.right)
-                - (styleSheet.contentStackViewLayoutMarginsRelativeArrangement
-                    ? styleSheet.contentLayoutMargins.horizontalTotal
+                - (styleSheet.content.isLayoutMarginsRelativeArrangement
+                    ? styleSheet.content.layoutMargins.horizontalTotal
                     : 0)
                 - imageWidthPlusSpacing(measuring: image, styleSheet: styleSheet)
                 - detailWidthPlusSpacing
@@ -266,7 +266,7 @@ private extension Component.TitledDescription {
             }
         }
 
-        return min(width, maxImageWidth) + styleSheet.horizontalSpacingBetweenElements
+        return min(width, maxImageWidth) + styleSheet.content.spacing
     }
 }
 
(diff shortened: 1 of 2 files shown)


StyleSheets/StyleSheets/StackViewStyleSheet.swift
```diff
@@ -1,31 +1,36 @@
 import UIKit
 
-open class StackViewStyleSheet: ViewStyleSheet<UIStackView> {
+open class StackViewStyleSheet<View: UIStackView>: ViewStyleSheet<View> {
```
sergdort: I feel like we would need to add generic attribute to the all of the stylesheets for consumers to be able to subclass. As people can subclass views but if they subclass stylesheet they will not be able to provide their view.

Let's say I create subclass `MyLabel: UILabel` then I want to have `MyLabelStyleSheet: LabelStyleSheet<MyLabel>` so then in `apply(to view: MyLabel)` but now I can not do that.

sergdort:
> Things that shouldn't be customisable are now made customisable 

Yeah, I was thinking about it, but I feel like it's not a big deal it's essentially a programmer error, which is very easy to notice. (It essentially the same if you would set a wrong property to a view)

But benefits that this approach provide is much more, we do not need to replicate all properties of the `ViewStyleSheet` + most of properties of the `StackViewStylesheet` that we allow to customise

sergdort:
I do not have a strong opinion on this one, just thought that would be kind of wasteful to create a new Stylesheet because we do not want to allow customisation of axis. ¯\_(ツ)_/¯

sergdort:
Ok, fair enough 👍 Gonna make a subclass of `ViewStyleSheet` 👌

ilyapuchka:
As discussed in slack, we probably should use `InteractiveStyleSheet` for the content so that we can configure its highlighting color

sergdort:
I guess `TitledDescription.Stylesheet` should be interactive not the content as its view is `InteractiveView`

mluisbrown:
> if tomorrow we change TitledDescription to not use any stack view that would affect every single consumer because they will be using a stack view stylesheet to configure this component

@dmcrodrigues isn't this still the case with the content stylesheet?

dmcrodrigues:
@sergdort can you include the typealiases to avoid referring directly to stack view types?

```swift
typealias Distribution = UIStackView.Distribution
typealias Alignment = UIStackView.Alignment
```

BentoKit/BentoKit/Components/TitledDescription.swift
```diff
@@ -508,6 +508,35 @@ private extension Component.TitledDescription.View {
...
+            element.distribution = distribution
+            element.alignment = alignment
+            element.isLayoutMarginsRelativeArrangement = isLayoutMarginsRelativeArrangement
+            element.cornerRadius = cornerRadius
+            element.borderColor = borderColor?.cgColor
+            element.borderWidth = borderWidth
+        }
+    }
 
     public final class StyleSheet: BaseViewStyleSheet<View> {
```
ilyapuchka: Can you please make it a subclass of interactive view stylesheet as discussed yesterday?

BentoKit/BentoKit/Components/TitledDescription.swift
```diff
@@ -508,6 +508,35 @@ private extension Component.TitledDescription.View {
 }
 
 public extension Component.TitledDescription {
+    public final class ContentStyleSheet: ViewStyleSheet<BaseStackView> {
```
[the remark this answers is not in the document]
andersio: This doesn't need to be `InteractiveViewStyleSheet`. In TD specifically, the highlighting gesture used is the one of the root view, and you can choose where the highlighting is applied to (root or content).

BentoKit/BentoKit/Components/TitledDescription.swift
```diff
@@ -508,26 +508,50 @@ private extension Component.TitledDescription.View {
...
+        public var spacing: CGFloat
+        public var distribution: UIStackView.Distribution
+        public var alignment: UIStackView.Alignment
+        public var isLayoutMarginsRelativeArrangement: Bool
 
-    public final class StyleSheet: BaseViewStyleSheet<View> {
+        public init(
+            spacing: CGFloat = 16,
+            distribution: UIStackView.Distribution = .fill,
```
andersio: I don't think distributions should be customisable. Plenty of them violates the algorithm in the height computer.

BentoKit/BentoKit/Components/TitledDescription.swift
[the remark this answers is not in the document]
ilyapuchka: My concern was from API point of view will it make sense to use the highlighting colour from root style sheet when highlighting target is `content`, or it would better to use highlighting colour from the content style sheet.

BentoKit/BentoKit/Components/TitledDescription.swift
sergdort: It was customisable before that's why I left it

sergdort: Ah, no you're right it was not used before

PR #77, opened 2018-11-20
Added accessory view stylesheet to TitledDescription

ilyapuchka:
Currently there is no way to adjust the stylesheet of the accessory view, so it can for example have unneeded default highlight colour like this:

simulator screen shot - iphone 5s - 2018-11-20 at 15 15 26


andersio:
Potentially you could also extend `HighlightingGesture` to manipulate alpha instead of background color in this case. But this would be a very niche use & a minor touch on the UI.

ilyapuchka:
@andersio the final result is supposed to be like this. I think it will be better to change a background colour to give a feedback if we want, but at the same time the result of the action in this particular case will be very prominent (a call alert 😄)

screen shot 2018-11-20 at 23 13 13

andersio:
Cool!

PR #79, opened 2018-11-21
Add line break mode for Button and Label stylesheets

sergdort:
### Description

Add ability to customise `lineBreakMode` for `UILabel` and `UIButton`


sergdort:
Good point 👍

Issue #82, opened 2018-11-23
TitledDescription does not Respect number of line when calculating height


sergdort:
If you take a look on the screenshot, There are two labels, one of which has limit to 2lines, but component does not takes this into account when it calculates the height

![screenshot 2018-11-23 at 15 25 10](https://user-images.githubusercontent.com/4622322/48950743-1b4a6b80-ef34-11e8-9047-0f7de8055812.png)

andersio:
#83 has fixed this.

PR #83, opened 2018-11-27
Use LayoutManager to calculate height of UILabel its height

sergdort:
### Description

We have an issue with [`LabelStylesheet`](https://github.com/Babylonpartners/Bento/issues/82) where it does not respect `numberOfLines` to calculate the height of the label

### Solution 

Use prototype label to calculate its height


andersio:
If we move to use prototype cell for sizing, we should make everything be computed consistently. The implementations should be moved to `TextBoundComputing`, and there is no point to have two overloads for `String` and `NSString`.

sergdort:
@andersio I think we can't move implementation to `TextBoundComputing` because in this case it uses `UILabel` prototype, in another it may use something else e.g UITextView

sergdort:
@andersio Addressed your comments

StyleSheets/StyleSheets/LabelStyleSheet.swift
```diff
@@ -15,20 +14,20 @@ open class LabelStyleSheet: ViewStyleSheet<UILabel>, TextBoundComputing {
...
+
     public init(
         backgroundColor: UIColor? = .clear,
         font: UIFont = UIFont.preferredFont(forTextStyle: .body),
         textColor: UIColor = .black,
         textAlignment: TextAlignment = .leading,
         numberOfLines: Int = 0,
-        lineBrealMode: NSLineBreakMode = .byTruncatingTail
-        ) {
+        lineBreakMode: NSLineBreakMode = .byTruncatingTail
```
diegopetrucci: 👌

PR #85, opened 2018-11-30
Enable customisation of `TextInput`'s text field

dmcrodrigues:
### Description

We should be able to customise the font, text color and alignment of text when using a `TextInput` component.

### DIFF
--- a/StyleSheets/StyleSheets/TextStyleSheet.swift
+++ b/StyleSheets/StyleSheets/TextStyleSheet.swift
@@ -1,0 +1,56 @@
+import UIKit
+
+public protocol TextViewProtocol: AnyObject {
+    var font: UIFont? { get set }
+    var textColor: UIColor? { get set }
+    var textAlignment: NSTextAlignment { get set }
+}
+
+public protocol TextViewConfiguratorProtocol: AnyObject {
+    func set(font: UIFont, textColor: UIColor, textAlignment: TextAlignment)
+}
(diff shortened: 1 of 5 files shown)


StyleSheets/StyleSheets/TextStyleSheet.swift
```diff
@@ -0,0 +1,56 @@
+import UIKit
+
+public protocol TextViewProtocol: AnyObject {
```
andersio: You could say `where Self: UIView` as part of the protocol.

dmcrodrigues: Not sure exactly why but I'm getting this strange errors 😕

image

sergdort: This is because it does not know it's a reference type, you either need to make a method `mutating` or make a protocol `AnyObject where Self: TextViewProtocol`

sergdort (approved):

StyleSheets/StyleSheets/TextStyleSheet.swift
dmcrodrigues: But it's already defined in the definition of `TextViewConfiguratorProtocol`.

```swift
protocol TextViewConfiguratorProtocol: AnyObject
```

diegopetrucci (approved):

PR #86, opened 2018-12-11
Add behaviours to the components

sergdort:
## Description

At the moment we have several protocols that can provide additional behavior to the `Renderable` components

- `Deletable` allows for a component to be deleted using swipe to delete action when it is present in the `UITableView`
- `ComponentLifecycleAware` allows for a component to be notified when its view lifecycle events happen e.g `willDisplayItem` `didEndDisplayingItem`
- `MenuItemsResponding` allows for a component to provide `[UIMenuItem]`  which will be presented in the contextual menu when a user long presses on the item in the `UITableView` or `UICollectionView`.

At the moment we achieve this by conforming each component to these protocols, when we need this behavior, which leads to:

- Every time in our codebase we encounter a situation where we need one of these behaviors we end up creating PR to the `BentoKit` library.
- This makes our components implementation bigger and bigger, essentially the same issue what `UIKit` has

## Proposal

Instead of using inheritance (in this case implementation inheritance) let's use composition via `Decorator` pattern + "operators" similar we have in `RAS` (e.g every time we what to alter the behaviour of a sequence we just apply operator)

Example:
Let's say we want to make a component to be a `Deletable` instead of conforming it to the `Deletable` we can just apply `Deletable` behavior to it:

```swift
let section = Section(id: 0)
            |---+ Node(id: 0, component: Component()
                .deletable(deleteActionText: "Delete", didDelete: {
                	// Do something about it
                })
            )
```

So the pros of this approach are:

-  all the logic related to the `Deletable` behavior is now in 2 places the `DeletableComponent` and in the `Adapter` 
- we can now make any component `Deletable` without changing the source code which aligns with Open for Extention closed for modification principle
- makes our components smaller and be focused on one thing which is rendering

And IMO this is another good example why **Composition** > **Inheritance**

## ⚠️

This is a breaking change to our code base but compiler will help use 🙈


Bento/Renderable/AnyRenderable.swift
```diff
@@ -99,13 +106,16 @@ private class AnyRenderableBox<Base: Renderable>: AnyRenderableBoxBase where Bas
...
 
-private class AnyRenderableBoxBase {
+class AnyRenderableBoxBase {
     var reuseIdentifier: String { fatalError() }
 
     var viewType: Any.Type { fatalError() }
 
     init() {}
 
+    func asAnyRenderable() -> AnyRenderable {
```
andersio: there doesn't seem any strong reason of not using the initialiser directly.

Bento/Renderable/Deletable.swift
```diff
@@ -1,15 +1,39 @@
...
     func delete()
 }
 
-public extension Deletable {
-    var backgroundColor: UIColor? {
-        return nil
+final class DeletableComponent<Base: Renderable>: AnyRenderableBox<Base>, Deletable where Base.View: UIView  {
+    let deleteActionText: String
+    let backgroundColor: UIColor?
+    private let source: AnyRenderableBox<Base>
```
andersio: You have the real `Base` here. Why is it necessary to type erase it?

Bento/Renderable/ComponentLifecycleAware.swift
```diff
@@ -7,3 +7,35 @@ public protocol ViewLifecycleAware {
     func willDisplayView()
     func didEndDisplayingView()
 }
+
+final class LifecycleComponent<Base: Renderable>: AnyRenderableBox<Base>, ComponentLifecycleAware where Base.View: UIView {
+    private let source: AnyRenderableBox<Base>
```
andersio: Ditto

Bento/Renderable/Deletable.swift
sergdort: Because `Base` does not know how to search for behaviour. Essentially the composition of behaviours is like a liked list, when we ask to `cast(to: )` we first check if type is the one we are implementing, if not we propagate search up in composition hierarchy. For more details please see line 29 in this file

Bento/Renderable/AnyRenderable.swift
sergdort: This is just to enable chaining calls 
```swift
DeletableComponent(
	            source: self,
	            deleteActionText: deleteActionText,
	            backgroundColor: backgroundColor,
	            didDelete: didDelete
	        )
.asAnyRenderable()
```

Bento/Renderable/ComponentLifecycleAware.swift
```diff
@@ -7,3 +7,35 @@ public protocol ViewLifecycleAware {
...
+    private let _willDisplayView: (() -> Void)?
```
dmcrodrigues: We don't use `_` as prefix and there's no collision between `willDisplayItem` and `willDisplayView`.

sergdort:
> Right now any behaviour requires to be part of BentoKit because consumers don't have access to AnyRenderableBox, shall we make it public to allow extension?

@dmcrodrigues  Yeah this is one of the issues I was thinking as well. And I think it would be nice for consumers of the library to be able to define their own behaviours, just like we can define our own operators with in RAS

sergdort:
@dmcrodrigues 

> Right now TableViewAdapter is coupled with Deletable behaviour but consumers may want to have a different implementation for example, shall we notify about this calls upstream so any component could observe them and implement any behaviour wanted/needed based on them? I think that would make it more flexible in the end.

My assumption was that if consumers would want to modify the behaviour, they would just need to override `tableView:trailingSwipeActionsConfigurationForRowAtIndexPath:`. Or am I missing something?

sergdort:
Oh, I see. 

So do you think we should just have something like 

```
protocol Swipeable {
    func swipeConfiguration() -> UISwipeActionsConfiguration
}
```

So in this case we could also support more then just `Delete` action.

sergdort:
But this would require to drop `iOS10` support, which is fine IMO ¯\_(ツ)_/¯

sergdort:
@dmcrodrigues regarding `AnyRenderableBox`. I think, if we are going to make it `public` we should reconsider the name, because it may confuse people as we have a `Box` :)

sergdort:
Also to allow custom behaviour we need an access to the adapter. So maybe we should provide an option to initialize `BoxViewController` with custom adapter type, what do you guys think?

cc @dmcrodrigues @andersio

dmcrodrigues:
> So do you think we should just have something like
> ```
> protocol Swipeable {
>    func swipeConfiguration() -> UISwipeActionsConfiguration
> }
> ```
> So in this case we could also support more then just Delete action.

Exactly but I think we need to build an abstraction over `UISwipeActionsConfiguration` and `UITableViewRowAction` because those are tied to `UITableView` and ideally we would like to have something sufficiently generic to be used interchangeably between table views and collection views.

> regarding AnyRenderableBox. I think, if we are going to make it public we should reconsider the name, because it may confuse people as we have a Box :)

I agree completely.

> Also to allow custom behaviour we need an access to the adapter. So maybe we should provide an option to initialize BoxViewController with custom adapter type, what do you guys think?

I think eventually this makes sense to give an extra bit of flexibility for more specific use cases but maybe we can hold off that change for now because we will need to "open" BoxTableViewAdapter, and ideally I'd like to enable custom behaviours without touching in any adapter the same we do with ReactiveSwift as you mentioned.

sergdort:
@dmcrodrigues @andersio  Also I've figured that `Deletable` has a very important behaviour (because it's destructive), if you take a look [here](https://github.com/Babylonpartners/Bento/blob/master/Bento/Adapters/TableViewAdapter.swift#L164) we are notifying component after animation completes so we do not get any UI glitches

So I would like to propose to leave it as it is and once we add an obstruction on `UISwipeActionsConfiguration` we may add support for additional swipe actions (which are not destructive)

dmcrodrigues:
👍 I'd like to take a look on that but for now I'm happy in merging this as it is.

Issue #80, opened 2018-11-21
Enable Size Caching by default


andersio:
## Problem Statement
Enable size caching by default in Bento, for both UITableView and UICollectionView (if using UICollectionViewFlowLayout).

This helps scrolling performance for use cases with a low frequency of changes & a moderate amount of screen complexity.

The problem is, however, the incurred cost during re-rendering of new `Box`s. Bento must invalidate all cached sizes before starting to render any new `Box`, unless the component does not change at a given ID path. Pragmatically speaking, this cannot happen for a majority of components (e.g. BentoKit) falling back to the pointer equality (conservative but correct), due to them carrying non-equatable properties e.g. callback.

So if we enable size caching without addressing this issue, for screens that have high burst of state changes e.g. reacting to continuous text input, the cost of precomputation would be incurred **every time** a new `Box` is rendered, which is incredibly unideal.

## Synthetic Benchmark

`TitledDescription`, 1000 iterations, -O -wholemodule

| Item | Result |
| --- | --- |
| Full equality of layout affecting properties* | Small Strings:  sub 1 ms.<br>5000+ character strings: sub 2 ms  |
| Custom Height Computation |  sub 100 ms  |
| AutoLayout height computation  | > 1000 ms |

\* Style sheets & RAS properties are compared by value.

## Observation

Evaluating equality of a subset of component properties is considerably cheaper (2-3 order of magnitude) than computing the actual layout height.

In order words, it is beneficial and pragmatic to evaluate equality so as to avoid as much layout computation as possible, lowering the performance impact of sizing ahead of time.

## Proposed optimization

- Introduce a notion of “component layout equivalence” to replace component equation.    

   When we say `C.isLayoutEquivalent(a, b) == true`, all properties contributing to the layout from both instances are equal. In other words, rendering `a` and `b` should result in the same exact layout.

   `Renderable` would however provide a `false` returning default implementation. This effectively replicates the current behaviour (correct but costly), while being (mostly) source compatible.

    BentoKit components should provide implementations for this requirement.

- Remove `Equatable` inheritance from `Renderable`.

   We should not reappropriate `==` for component layout equivalence, since it would violate the substitutability requirement mandated by the Equatable contract.

- With the above two measures in place, Bento may start pre-compute component height using AutoLayout **by default** .

   Bento should invalidate only cached size for ID paths failing the component layout equivalence test.

- (Optionally) Removing `HeightCustomizing`.

[...]

PR #88, opened 2018-12-21
Composable StyleSheet

andersio:
Playbook Proposal: https://github.com/Babylonpartners/ios-playbook/pull/35

Credit: The Winter FP workshop & @dmcrodrigues, @ilyapuchka and @stephencelis.

## Problem Statement
The current inheritance based StyleSheet approach is unideal in three areas:

1. Ease of Use: Requiring upfront declarations of all styleable properties.

    We would have to constantly match UIKit changes. Custom views would require manual implementation.

2. Mental Model: Awkwardness when being applied to views that are opened for extension.

    When one defines a non generic `ExtensibleViewStyleSheet` for an `ExtensibleView`, the `StyleSheetProtocol.Element` associated type is immediately satisfied with the concrete view type. This means in a generic environment bound to `StyleSheetProtocol`, only `apply(to _: ExtensibleView)` would be recognised since `Element == ExtensibleView` per the conformance. In other words, all stylesheet subclasses must override this base class typed method & are required to downcast at runtime to access stuff defined in the subclass.

    On the other hand, when one defines a generic `ExtensibleViewStyleSheet<T: ExtensibleView>` stylesheet, it does address the said issue by satisfying `StyleSheetProtocol.Element` with a generic parameter. However, it leads to awkward and verbose spellings at use sites in general, e.g. `ExtensibleViewStyleSheet<ExtensibleView>`.

3. Correctness: Unintended Sharing

    Stylesheets are currently classes so as to be able to take advantage of inheritance, and this makes them prone to unintended sharing. With Swift offering first-class value type, this might amplify its possibility due to common perception of configurations being modelled as value types.

## Identified Requirements

1. The approach must be applicable to reusable views.

    When multiple style sheets with symmetrical differences are applied in a row in any order, the reused view would not end up in an indeterministic state.

1. The approach must be compatible with view subclassing.

1. The approach should address unintended sharing if possible.

## Proposed Solution

A new value type `StyleSheet<View>` shall be introduced, with the following features:

1. Copy on write & value semantics: Backed by CoW `Dictionary`.

2. Built around Key Paths introduced since Swift 4.

3. Inverses are produced for deterministic use on reusable views.

    Before applying a stylesheet, a snapshot of all affected key paths is taken, and is returned to the callee after the application. It is up to the callee whether to and when to apply the snapshot for change reversal.

    In the case of a reusable view/cell, the expectation is that the view would store the inverse/snapshot & apply it as part of the cleanup e.g. in `prepareForReuse()`.
    
    This eliminates the state indeterminism of closure-based styling approaches, while not requiring any default value to be declared upfront like how our `StyleSheet` framework currently does. 

4. Can be used for any view and any arbitrary subclass without type declaration.

5. `Equatable` conformance as a bonus, useful for #80 in determining layout equivalence.

### Potential Issues

Not all stying parameters are exposed as instance properties.

`UIButton` is the most apparent example. We could mitigate this by providing extensions with computed properties of `[UIControl.State: U]`.

    
## Impact on existing APIs
The value type `StyleSheet` is an additive change alone.

However, assuming we subsequently replace all existing BentoKit component style sheets, this shall be a massive source breaking change.

### Potential Migration Path
While script-based and staged migration is a viable solution, additional effort on top of snapshot testing might be required to verify that the removal of style sheet defaults does not lead to changes in visual appearance.

### DIFF
--- a/Bento.xcodeproj/project.pbxproj
+++ b/Bento.xcodeproj/project.pbxproj
@@ -55,6 +55,8 @@
 		58FC4449207CFBD700DA3614 /* IconTitleDetailsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 58FC443F207CFBD700DA3614 /* IconTitleDetailsView.swift */; };
 		5B02B99322503A370089371A /* AdapterProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B02B99122503A370089371A /* AdapterProtocol.swift */; };
 		5B02B99422503A370089371A /* AdapterStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B02B99222503A370089371A /* AdapterStore.swift */; };
+		5B2CA2A922692E460046D712 /* StyleSheetProtocol.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B2CA2A822692E460046D712 /* StyleSheetProtocol.swift */; };
+		5B2CA2AB22692E870046D712 /* LegacyStyleSheetTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B2CA2AA22692E870046D712 /* LegacyStyleSheetTests.swift */; };
 		5B6807942266818F006AA479 /* SizeCachingTableViewTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5B6807932266818F006AA479 /* SizeCachingTableViewTests.swift */; };
 		61B40919208523F60063DE25 /* FoodListViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61B40916208523F50063DE25 /* FoodListViewModel.swift */; };
 		61B64BB12086561B0092082C /* FoodListRenderer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 61B64BB02086561B0092082C /* FoodListRenderer.swift */; };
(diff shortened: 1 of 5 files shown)


StyleSheets/StyleSheets/StyleSheet.swift
```diff
@@ -1,29 +1,95 @@
...
-    public func compose<T>(_ property: ReferenceWritableKeyPath<Self, T>, _ value: T) -> Self {
-        self[keyPath: property] = value
-        return self
+    public func apply(to view: inout View) -> Inverse {
+        defer {
+            nodes.forEach { $0.value.apply(to: &view) }
+        }
+        
+        return StyleSheet().with {
+            $0.nodes = self.nodes.mapValues { $0.current(in: view) }
```
ilyapuchka: I suppose this will only work if a property is changed only once after the view is created or reused (when all changes would be reverted to original values), so the framework should either ensure that, or it should store all inverse stylesheets and apply them in reversed order when preparing view for reuse.

ilyapuchka: @andersio what I mean is that if when configuring view after it was reused (and has value of `value0`) one sets the same property twice, first to `value1` and then to `value2`, on reuse it will be reset to the `value1`, not to `value0`

andersio: @ilyapuchka Latest fixes have tackled this issue with unit tests attached. It also covers changes at partially overlapping key paths, e.g. `\.layoutMargins.left` and `\.layoutMargins` in any order.

StyleSheets/StyleSheets/StyleSheet.swift
```diff
@@ -1,29 +1,152 @@
...
+    /// Retrieve the value at the specific key path.
+    public func value<Value: Equatable>(for keyPath: WritableKeyPath<View, Value>) -> Value? {
+        let key = EntryKey(keyPath)
+        return entries[key] as! Value?
+    }
+
+    public mutating func set<Value: Equatable>(_ keyPath: WritableKeyPath<View, Value>, _ newValue: Value) {
+        let key = EntryKey(keyPath)
+        entries[key] = newValue
+        orderedKeys.append(key)
```
inamiy: Probably need to consider about multiple same `keyPath` being set.

inamiy: Please add test case for this scenario as well.
(In the meeting, we conluded this as "not inverting two values with same key, but only once")

andersio: Yeah. I was wrong in the meeting, since I forgot about having made it to record insertion order of the keys. 😅

PR #92, opened 2019-01-10
Migrate to Swift 4.2

NSMyself:
I was toying around with Bento in Xcode 10.1 and noticed that Bento's still using Swift 4.0 so I decided to open this PR. It was pretty straightforward but:

1. I didn't update the dependencies (not sure if you want this or not right now), so there are still warnings popping up regarding them.

2. there was a `different optionality` warning in the CollectionViewAdapter due to the fact that the method ```open func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath?, withSender sender: Any?) -> Bool``` is using an optional IndexPath but the actual UICollectionViewDelegate wants it to be non-optional. 

Guessing that's like that because of Objective-C interop? Ideally, the indexPath shouldn't really be an optional.

### DIFF
--- a/Bento/Adapters/CollectionViewAdapter.swift
+++ b/Bento/Adapters/CollectionViewAdapter.swift
@@ -114,11 +114,15 @@ open class CollectionViewAdapterBase<SectionID: Hashable, ItemID: Hashable>
 
     @objc(collectionView:canPerformAction:forItemAtIndexPath:withSender:)
     open func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath?, withSender sender: Any?) -> Bool {
-        guard let indexPath = indexPath,
-            let component = sections[indexPath.section].items[indexPath.row].component(as: MenuItemsResponding.self) else {
-            return false
+        guard let indexPath = indexPath else { return false }
+        return self.collectionView(collectionView, canPerformAction: action, forItemAt: indexPath, withSender: sender)
+    }
+    
+    open func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath, withSender sender: Any?) -> Bool {
+        guard let component = sections[indexPath.section].items[indexPath.row].component(as: MenuItemsResponding.self) else {
+                return false
         }
-
+        
         return component.responds(to: action)
     }
 
--- a/Bento.xcodeproj/project.pbxproj
+++ b/Bento.xcodeproj/project.pbxproj
@@ -526,6 +526,7 @@
 				TargetAttributes = {
 					5829D264200FB092001E020D = {
 						CreatedOnToolsVersion = 9.2;
+						LastSwiftMigration = 1010;
 						ProvisioningStyle = Automatic;
 					};
 					58BA755B2016303B0050D5F1 = {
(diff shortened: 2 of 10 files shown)


mluisbrown:
> here was a different optionality warning in the CollectionViewAdapter due to the fact that the method open func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath?, withSender sender: Any?) -> Bool is using an optional IndexPath but the actual UICollectionViewDelegate wants it to be non-optional.

The `UICollectionViewDelegate` implementation is [this one](https://github.com/Babylonpartners/Bento/pull/92/files#diff-e2b10cfde1e83638aaf367d2ce45e33dR115) with the `@objc` attribute. The other one has a non optional `IndexPath` and is only called from the previous one if the `IndexPath` is not nil, so you can ignore the warning.

NSMyself:
Yeah sure, I'll do so either tonight or tomorrow

mluisbrown:
BentoKit and Stylesheets are already on 4.2 🤦‍♂️

andersio:
Flipping PR status to trigger CI.

Bento/Adapters/CollectionViewAdapter.swift
```diff
@@ -114,11 +114,15 @@ open class CollectionViewAdapterBase<SectionID: Hashable, ItemID: Hashable>
...
     @objc(collectionView:canPerformAction:forItemAtIndexPath:withSender:)
     open func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath?, withSender sender: Any?) -> Bool {
-        guard let indexPath = indexPath,
-            let component = sections[indexPath.section].items[indexPath.row].component(as: MenuItemsResponding.self) else {
-            return false
+        guard let indexPath = indexPath else { return false }
+        return self.collectionView(collectionView, canPerformAction: action, forItemAt: indexPath, withSender: sender)
+    }
+    
+    open func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath, withSender sender: Any?) -> Bool {
```
mluisbrown: Why not just change `IndexPath?` to `IndexPath` and remove the guard for it, instead of adding a new method?

Bento/Adapters/CollectionViewAdapter.swift
```diff
@@ -114,11 +114,15 @@ open class CollectionViewAdapterBase<SectionID: Hashable, ItemID: Hashable>
 
     @objc(collectionView:canPerformAction:forItemAtIndexPath:withSender:)
     open func collectionView(_ collectionView: UICollectionView, canPerformAction action: Selector, forItemAt indexPath: IndexPath?, withSender sender: Any?) -> Bool {
```
andersio: This should be the one satisfying the conformance since we introduced this to fix a UIKit bug. 😬

Bento/Adapters/CollectionViewAdapter.swift
mluisbrown: Ok, looks like `UIKit` can pass `null` to params that are annotated as non-null 🤦‍♀️

andersio:
I guess we should leave a `// NOTE:` comment there stating this.

Bento/Adapters/CollectionViewAdapter.swift
mluisbrown: Does that mean that the introduction of the other one is now satisfying the conformance instead?

Bento/Adapters/CollectionViewAdapter.swift
NSMyself: I saw that @sergdort made a commit specifically for nil IndexPaths last November and, as such, I chose to maintain that same logic

Bento/Adapters/CollectionViewAdapter.swift
NSMyself: @andersio out of curiosity, is there a Rdar for that UIKit bug? Only managed to find this [one](https://openradar.appspot.com/31375101)

PR #93, opened 2019-01-14
Screen Lifecycle.

andersio:
Introduce a mechanism for `BoxViewModel` to be notified of the screen lifecycle.

`BoxViewModel` conforming to `ScreenLifecycleAware` would be invoked with `send(_: ScreenLifecycleEvent)` whenever the status of the screen changes. It is up to the view model whether or not it responds to the lifecycle event with a state transition.

`ScreenLifecycleEvent` exposes stages of presentation and dismissal, and special attributes for knowing whether the screen is initially presented or is to be removed indefinitely.

### DIFF
--- a/BentoKit/BentoKit/Screen/BoxViewController.swift
+++ b/BentoKit/BentoKit/Screen/BoxViewController.swift
@@ -42,6 +42,22 @@ open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, App
         }
     }
 
+    private var lifetimeObserver: ScreenLifecycleObserving? {
+        return viewModel as? ScreenLifecycleObserving
+    }
+
+    private var isPresentedInitially: Bool {
+        return (parent != nil && isMovingToParent)
+            || (presentingViewController != nil && isBeingPresented)
+            || (parent == nil && presentingViewController == nil)
+    }
+
+    private var isRemovedPermanently: Bool {
+        return (parent != nil && isMovingFromParent)
+            || (presentingViewController != nil && isBeingDismissed)
+            || (parent == nil && presentingViewController == nil)
+    }
+
     public init(viewModel: ViewModel,
                 renderer: Renderer.Type,
                 rendererConfig: Renderer.Config,
@@ -66,6 +82,8 @@ open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, App
         super.viewDidLoad()
         setupTableViews()
         setupLayout()
+
+        lifetimeObserver?.send(.didLoad)
     }
 
     open override func viewWillAppear(_ animated: Bool) {
@@ -78,11 +96,15 @@ open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, App
 
             bindViewModel()
         }
+
+        lifetimeObserver?.send(.willAppear(isPresentedInitially: isPresentedInitially))
     }
 
     open override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         hasViewAppeared.value = true
+
+        lifetimeObserver?.send(.didAppear(isPresentedInitially: isPresentedInitially))
 
         keyboardChangeDisposable = NotificationCenter.default.reactive
             .keyboard(.willChangeFrame)
@@ -117,11 +139,16 @@ open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, App
     open override func viewWillDisappear(_ animated: Bool) {
         super.viewWillDisappear(animated)
         hasViewAppeared.value = false
+
+        lifetimeObserver?.send(.willDisappear(isRemovedPermanently: isRemovedPermanently))
+
     }
 
     open override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         keyboardChangeDisposable?.dispose()
+
+        lifetimeObserver?.send(.didDisappear(isRemovedPermanently: isRemovedPermanently))
     }
 
     open override func viewDidLayoutSubviews() {
--- a/BentoKit/BentoKit/Screen/BoxViewController.swift
+++ b/BentoKit/BentoKit/Screen/BoxViewController.swift
@@ -42,10 +42,6 @@ open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, App
         }
     }
 
-    private var lifetimeObserver: ScreenLifecycleObserving? {
-        return viewModel as? ScreenLifecycleObserving
-    }
-
     private var isPresentedInitially: Bool {
         return (parent != nil && isMovingToParent)
             || (presentingViewController != nil && isBeingPresented)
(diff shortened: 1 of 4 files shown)


BentoKit/BentoKit/Screen/ScreenLifecycleEvent.swift
```diff
@@ -0,0 +1,35 @@
+public protocol ScreenLifecycleObserving {
+    func send(_ event: ScreenLifecycleEvent)
+}
+
+public struct ScreenLifecycleEvent {
+    public let status: Status
+    public let isBeingPresentedInitially: Bool
+    public let isBeingRemovedIndefinitely: Bool
```
mluisbrown: As mentioned in my review of https://github.com/Babylonpartners/ios-playbook/pull/22 IMO we should change `Indefinitely` to `Permanently` for clarity (using a non-negative term).

mluisbrown (changes requested):
Please change `Indefinitely` to `Permanently`. Looks great otherwise 👍

BentoKit/BentoKit/Screen/ScreenLifecycleEvent.swift
```diff
@@ -0,0 +1,35 @@
...
+
+    public init(status: Status,
+                isBeingPresentedInitially: Bool,
+                isBeingRemovedIndefinitely: Bool) {
+        self.status = status
+        self.isBeingPresentedInitially = isBeingPresentedInitially
+        self.isBeingRemovedIndefinitely = isBeingRemovedIndefinitely
+    }
+
+    public enum Status {
```
sergdort: I wonder if it's enough just to have it as an `enum` with associated values? Instead of `Status + Event`

mluisbrown: 👍 especially as the `isBeing...` flags don't apply to each situation.

sergdort: Having `viewDidLoad` would be nice as well

BentoKit/BentoKit/Screen/ScreenLifecycleEvent.swift
```diff
@@ -0,0 +1,35 @@
+public protocol ScreenLifecycleObserving {
+    func send(_ event: ScreenLifecycleEvent)
+}
+
+public struct ScreenLifecycleEvent {
+    public let status: Status
+    public let isBeingPresentedInitially: Bool
```
sergdort: I guess this would not be necessary if we have `viewDidLoad`.

sergdort (changes requested):
I think we need `viewIDidLoad` as well.
And it seems like it's better for it to be an enum with associated values.

BentoKit/BentoKit/Screen/ScreenLifecycleEvent.swift
andersio: Having `didLoad` does not change the fact that there are at least 8 different scenarios:

* will/did appear AND "is being presented initially"
* will/did disappear AND "is being removed permanently"
* will/did appear AND NOT "is being presented initially"
* will/did disappear AND NOT "is being removed permanently"


`viewDidLoad` may also happen way before `viewWillAppear`, and a view controller might be reused, so it might not necessarily be what one wants.

andersio:
`didLoad` in BoxViewController is effectively the same as doing side effect in the VM initialiser. `willAppear` semantically may happen way after `didLoad` (e.g. initialised but not being presented), and could happen more than once for any given instance.

BentoKit/BentoKit/Screen/ScreenLifecycleEvent.swift
sergdort: > and a view controller might be reused
Well I do not think it's a good practice. ¯\_(ツ)_/¯

sergdort:
Also @mluisbrown made a good point [here](https://github.com/Babylonpartners/Bento/pull/93#discussion_r247434544)

BentoKit/BentoKit/Screen/ScreenLifecycleEvent.swift
inamiy:
```swift
enum Event {
    case viewDidLoad
    case viewWillAppear(isInitial: Bool)
    case viewDidAppear(isInitial: Bool)
    case viewWillDisappear(isFinal: Bool)
    case viewWillDisappear(isFinal: Bool)
}
```

seems sufficient as already mentioned.

BentoKit/BentoKit/Screen/BoxViewController.swift
```diff
@@ -42,6 +42,22 @@ open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, App
...
     }
 
+    private var lifetimeObserver: ScreenLifecycleObserving? {
+        return viewModel as? ScreenLifecycleObserving
+    }
+
+    private var isBeingPresentedInitially: Bool {
+        return (parent != nil && isMovingToParent)
+            || (presentingViewController != nil && isBeingPresented)
+            || (parent == nil && presentingViewController == nil)
```
inamiy, line 48: It's good to have doc-comment for what each line means.

andersio:
Accepted playbook proposal: https://github.com/Babylonpartners/ios-playbook/blob/master/TechnicalDocuments/Proposals/ScreenLifecycleObservation.md

dmcrodrigues:
@andersio what's the status of this one?

BentoKit/BentoKit/Screen/ScreenLifecycleEvent.swift
```diff
@@ -0,0 +1,38 @@
+public protocol ScreenLifecycleObserving {
+    func send(_ event: ScreenLifecycleEvent)
```
inamiy: Do you have any thoughts about merging this protocol with `BoxViewModel` and its `func send(action: Action)` ?

It will be a bigger breaking change, but IMO having 2 separate protocols with 2 separate `func send` interfaces are redundant.

If we end up keeping this protocol, please add documentation for `ScreenLifecycleObserving` that requires ViewModel to be conformed alongside `BoxViewModel`.

andersio: I don't mind.

inamiy: ^ Above will look like this:

```swift
public protocol BoxViewModel: AnyObject {
    associatedtype State
    associatedtype Action

    var state: Property<State>  { get }

    func send(action: ViewLifeCycleOrUserAction<Action>)
}

enum ViewLifeCycleOrUserAction<Action> {
    case viewLifecycle(ScreenLifecycleEvent)
    case userAction(Action)
}
```

andersio: I prefer having them separate, like, they are from separated sources after all. Little point to glue them together artificially.

sergdort (approved):

mluisbrown (approved):

PR #94, opened 2019-01-17
Hide image and label. Add accessibility ids.

sergdort:
# Why 

After I refactored one of our screens from Forms to Bento. I've noticed some of the tests where failing.

# How

1. Add accessibility IDs to the ImageOrLabel view
2. Also I've added hide and show logic for the `label` and `imageView` based on the `content`.

# Note 

This may seem like quick fix, but I guess we need to have an accessibility id for this view anyway ¯\_(ツ)_/¯


andersio:
If we are adding a11y ID, shouldn't we at least allow users to specify a custom one?

sergdort:
Closing this for now

PR #99, opened 2019-01-30
Allow customizing Button type

### LINKED ISSUE #98: Component.Button stylesheet should inherit from InteractiveViewStyleSheet
The `Component.Button.Stylesheet` currently inherits from `BaseViewStyleSheet`, but the base view for the `Button` component is an `InteractiveView` so the stylesheet class should inherit from `InteractiveViewStyleSheet`.

A consequence of this bug is that you cannot set the `highlightColor` for a Button. That means that when you set custom `.highlighted` color for the button, it is overridden with the default `highlightColor` of the `InteractiveView`:

buttonbug

mluisbrown:
Also: Button should use InteractiveViewStyleSheet
Fixes #98 

Add a `type` to the Button component and allow this to be configured by the user. This allows using the `.custom` type, which is the only way to customize the button highlighted background colour without having a system fade applied. 

This was the actual cause of #98, not the incorrect style sheet type, but I fixed that also, just to be correct.

### DIFF
--- a/BentoKit/BentoKit/Components/Button.swift
+++ b/BentoKit/BentoKit/Components/Button.swift
@@ -65,6 +65,8 @@ extension Component.Button {
         private var huggingConstraints: [NSLayoutConstraint] = []
         private var strictConstraints: [NSLayoutConstraint] = []
 
+        fileprivate var interactionBehavior: InteractionBehavior = .becomeFirstResponder
+
         fileprivate var buttonType: UIButton.ButtonType = .system {
             didSet {
                 guard oldValue != buttonType else { return }
--- a/BentoKit/BentoKit/Components/Button.swift
+++ b/BentoKit/BentoKit/Components/Button.swift
@@ -7,6 +7,7 @@ extension Component {
 
         public let configurator: (View) -> Void
         public let styleSheet: StyleSheet
+        public let type: UIButton.ButtonType
 
         private let heightComputer: (CGFloat, UIEdgeInsets) -> CGFloat
 
@@ -15,8 +16,10 @@ extension Component {
             isEnabled: Bool = true,
             isLoading: Bool = false,
             didTap: (() -> Void)? = nil,
+            type: UIButton.ButtonType = .system,
             styleSheet: StyleSheet
         ) {
+            self.type = type
             self.configurator = { view in
                 view.isLoading = isLoading
                 view.button.isEnabled = isEnabled
@@ -39,6 +42,10 @@ extension Component {
             self.styleSheet = styleSheet
         }
 
+        public func generate() -> View {
+            return View(type: type)
+        }
+
         public func height(forWidth width: CGFloat, inheritedMargins: UIEdgeInsets) -> CGFloat {
             return heightComputer(width, inheritedMargins)
         }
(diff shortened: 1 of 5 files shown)


mluisbrown:
> the UIButton should be dynamically recreated whenever the button type changes which is also a technique already used in several BentoKit & internal components.

Do you have an example?

sergdort:
> using distinct reuse IDs
Yeah this should probably work

sergdort:
@mluisbrown you just need to override `reuseId` property on the component. This is why it convenient for it to be instance property, @andersio 😜

andersio:
To be clear, changing reuse identifier would solve the issue, provided that we also fix the adapters which currently does not handle component views with mismatching reuse identifiers as a result of silent updates (aka in-place reload). Just that I don't think it is a very compelling solution in _mitigating_ this class of issues, which is more about developer experience.

mluisbrown:
@andersio I changed it so the Button is always recreated if the type changes.

BentoKit/BentoKit/Components/Button.swift
```diff
@@ -61,25 +61,67 @@ extension Component.Button {
...
+                return _huggingConstraints
+            }
 
-        private lazy var strictConstraints: [NSLayoutConstraint] = [
-            button.leadingAnchor.constraint(equalTo: layoutMarginsGuide.leadingAnchor)
-                .withPriority(.defaultHigh),
-            layoutMarginsGuide.trailingAnchor.constraint(equalTo: button.trailingAnchor)
-                .withPriority(.defaultHigh)
-        ]
+            _huggingConstraints = [
```
andersio: I guess we should probably create these as part of `setupLayout()` now.

BentoKit/BentoKit/Components/Button.swift
```diff
@@ -61,25 +61,67 @@ extension Component.Button {
...
+                button = Button(type: buttonType)
+
+                _huggingConstraints = []
+                _strictConstraints = []
+
+                button.setContentHuggingPriority(.required, for: .vertical)
+                button.addTarget(self, action: #selector(buttonPressed), for: .touchUpInside)
+                setupLayout()
+
+                if hugsContent {
```
andersio: Shouldn't we activate the strict constraints if we do not want to hug content?

BentoKit/BentoKit/Components/Button.swift
```diff
@@ -61,25 +61,67 @@ extension Component.Button {
...
 
-        public let button = Button(type: .system).with {
-            $0.setContentHuggingPriority(.required, for: .vertical)
+        public var button = Button(type: .system)
+
+        fileprivate var buttonType: UIButton.ButtonType = .system {
+            didSet {
+                guard oldValue != buttonType else { return }
+
+                activityIndicator.removeFromSuperview()
```
andersio: The logic here can be incorporated into `setupLayout()`, so we just need to invoke `setupLayout()` here to reconfigure as needed. Would centralize all these relevant bits too.

mluisbrown:
@andersio comments addressed.

mluisbrown:
Comments addressed @andersio

PR #101, opened 2019-02-02
Layout Equivalence.

andersio:
As established in #80, the ultimate goal of Bento is to enable size caching by default, while minimising the performance hit for screens with high frequency changes (e.g. forms).

To achieve the latter, Bento needs to determine whether component instances result in equivalent layouts, and in turn avoid having to invalidate any cached layout information whose re-computation might take one or two order of magnitude longer in time.

This PR introduces a mean for the components to provide a definition of layout equivalence, as an optional `Renderable.layoutEquivalence(_:_:)` requirement.

~~**WIP:** Still thinking about how this should interact with behavior composition e.g. `Deletable` and stuff.~~

### DIFF
--- a/Bento/Renderable/LayoutCompatibility.swift
+++ b/Bento/Renderable/LayoutCompatibility.swift
@@ -8,5 +8,5 @@ public enum LayoutEquivalence {
 
     /// The component does not have a definition of layout equivalence. Bento must treat all instances as always
     /// different from each other.
-    case unknown
+    public static var unknown: LayoutEquivalence { return .different }
 }


Bento/Renderable/LayoutCompatibility.swift
```diff
@@ -0,0 +1,12 @@
...
+public enum LayoutEquivalence {
+    /// The layouts are the same, and Bento may reuse any cached layout parameter.
+    case same
+
+    /// The layouts should be different from each other, and Bento must invalidate any cached layout parameter.
+    case different
+
+    /// The component does not have a definition of layout equivalence. Bento must treat all instances as always
+    /// different from each other.
+    case unknown
```
sergdort: Why do we need this case if we treat it as `different`?

PR #109, opened 2019-02-13
Adds support for leading and trailing horizontal alignment.

mluisbrown:
This adds the `alignment` property to `Component.Button.StyleSheet` to allow having buttons aligned with the leading or trailing `layoutMargin`s of the component view.

Currently only center alignment is supported. The default continues to be `.center` and this makes the change backwards compatible with existing code.

Also fixes a bug so the height of a button which has only an image (and no text) is correct, when `enforcesMinimumHeight` is `false`.


TheAdamBorek:
It's backward compatible so there I think there's no reason to not merge it :)

mluisbrown:
All Babylon iOS tests passed with this change 👍

PR #113, opened 2019-02-21
Added contentView to TextInput

mluisbrown:
Added a `contentView` to the `TextInput` component to allow for having background colour that respect layout margins.

### DIFF
--- a/BentoKit/BentoKit/Components/TextInput.swift
+++ b/BentoKit/BentoKit/Components/TextInput.swift
@@ -65,7 +65,9 @@ extension Component.TextInput {
             case fillProportionally(CGFloat)
         }
 
-        private let container: UIStackView
+        fileprivate let contentView = UIView().with {
+            $0.translatesAutoresizingMaskIntoConstraints = false
+        }
 
         private var titleLabelWidthConstraint: NSLayoutConstraint? {
             willSet { titleLabelWidthConstraint?.isActive = false }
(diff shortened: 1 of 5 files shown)


BentoKit/BentoKit/Components/TextInput.swift
```diff
@@ -65,7 +65,9 @@ extension Component.TextInput {
             case fillProportionally(CGFloat)
         }
 
-        private let container: UIStackView
+        fileprivate let contentView = UIView().with {
+            $0.translatesAutoresizingMaskIntoConstraints = false
```
TheAdamBorek: [nit] this is not needed I think. You use
```swift
 .add(to: self)
                .pinEdges(to: layoutMarginsGuide)
```

which already has it

BentoKit/BentoKit/Components/TextInput.swift
```diff
@@ -95,21 +97,28 @@ extension Component.TextInput {
         var textDidChange: Optional<(String?) -> Void> = nil
 
         override init(frame: CGRect) {
+            super.init(frame: frame)
+            setupLayout()
+        }
+
+        private func setupLayout() {
+            preservesSuperviewLayoutMargins = true
```
TheAdamBorek: isn't this true by default?

mluisbrown: No, it's fase by default.

PR #115, opened 2019-02-25
Added support for a content view to MultilineTextInput

mluisbrown:
* Added support for `didChangeText` to allow keeping model state updated with the current text content, in order to allow correct text view sizing.

* Re-factored the StyleSheet to use `TextStyleSheet` and add extra configuration.

* Made the placeholderLabel adjust its position in line with `textView.textContainerInset`

* Added `showsFocusToolbar` to configure whether the focus toolbar is shown or not.


dmcrodrigues:
I have mix feelings around this PR for a couple of reasons:

- I'm not sure if this button would be that frequent, in our case it seems we only have a single usage for it for example, which I think make us questioning if it should be a feature to be embedded in `MultilineTextInput` or if we should kind of compose and reuse both `MultilineTextInput` and `Button` in a new component which displays both as we are doing right now;
- It seems we are tying `didFinishEditing` to the tapping of the button introduced which may be unexpected for consumers if they rely on that particular event to do some work unrelated with the activation of the button itself;
- `inputAccessoryView` is being defined based on the button being visible or not which is being managed based on the text view itself being empty or not, I feel we making multiple assumptions here and consumers cannot customise this if they need it to do so.

I don't oppose to this changes but I'm not sure if they are the right direction for something that should be completely generic and agnostic of any use case, I feel we are kind of tying it to some extent to a specific use case that we have in our product, especially in the last two points.

mluisbrown:
@dmcrodrigues: 

> I'm not sure if this button would be that frequent

IMO there are other potential use cases for a send button within this component. Also, if the button is not enabled, it has no effect on the component.

> It seems we are tying `didFinishEditing` to the tapping of the button introduced which may be unexpected for consumers

Currently tapping the `Done` button on the `inputAccessoryView` does the same thing, so the only difference is that having a send button assumes (IMO logically) that there is no input accessory view.

> `inputAccessoryView` is being defined based on the button being visible or not which is being managed based on the text view itself being empty or not,

It's only defined based on the send button being visible or not. The text view being empty or not is not a factor. The view being empty or not only affects whether the send button (if it's visible) is *enabled* or not.

andersio:
It is probably not that uncommon. Our appointment screen has instead an add photo button, for example.

BentoKit/BentoKit/Components/MultilineTextInput.swift
```diff
@@ -29,40 +40,73 @@ extension Component.MultilineTextInput {
...
             // can take precedence when the content height is being determined.
             NSLayoutConstraint.activate([
-                layoutMarginsGuide.topAnchor.constraint(equalTo: placeholderLabel.topAnchor),
-                layoutMarginsGuide.leadingAnchor.constraint(equalTo: placeholderLabel.leadingAnchor),
-                layoutMarginsGuide.trailingAnchor.constraint(equalTo: placeholderLabel.trailingAnchor),
-                layoutMarginsGuide.bottomAnchor.constraint(equalTo: placeholderLabel.bottomAnchor)
+                contentView.layoutMarginsGuide.topAnchor.constraint(equalTo: placeholderLabel.topAnchor),
+                contentView.layoutMarginsGuide.leadingAnchor.constraint(equalTo: placeholderLabel.leadingAnchor),
+                contentView.layoutMarginsGuide.trailingAnchor.constraint(equalTo: placeholderLabel.trailingAnchor),
+                contentView.layoutMarginsGuide.bottomAnchor.constraint(equalTo: placeholderLabel.bottomAnchor)
```
TheAdamBorek: [nit]
```swift
placeholderLabel.pinToEdges(contentView.layoutMargins)
```

BentoKit/BentoKit/Components/MultilineTextInput.swift
```diff
@@ -29,40 +40,73 @@ extension Component.MultilineTextInput {
...
             $0.textContainer.lineFragmentPadding = 0.0
         }
 
-        let placeholderLabel = UILabel()
-        var didFinishEditing: ((String) -> Void)?
+        fileprivate let placeholderLabel = UILabel().with {
+            $0.translatesAutoresizingMaskIntoConstraints = false
+        }
+        fileprivate let contentView = UIView()
+        fileprivate let sendButton = UIButton(type: .system).with {
```
TheAdamBorek: I don't know if `sendButton` is a correct name for component which should rather be generic.
I don't mind having the button but I think we can generalize naming here :)

sergdort:
I agree with @dmcrodrigues here, it seems to me that we are trying to impose our internal design on `BentoKit` which is suppose to be a collection of "default" UIKit components. 
This could lead to the problems that happens to `TitledDescription` interface, which is so complex, that people outside of our team are sometimes struggling to understand.

What I would suggest is maybe to compose `MultilineTextInput.View` with another by creating new component in our codebase (If you want to reuse the same behaviour). This may require to make some properties of `MultilineTextInput.View` public, which is fine ¯\_(ツ)_/¯

TheAdamBorek:
@sergdort That's also an option. On making properties public, I would even say that in BentoKit we should make properties public by default to allow View's composition

BentoKit/BentoKit/Components/MultilineTextInput.swift
mluisbrown: We can't use `pinToEdges` here because the bottom constraint has a lower priority (`.defaultHigh`)

BentoKit/BentoKit/Components/MultilineTextInput.swift
mluisbrown: Well, it's `send` because tapping the button will invoke `didFinishEditing` and `resignFirstResponder`.

mluisbrown:
@sergdort @dmcrodrigues the problem with using composition in this case is that I would need to "hijack" the `UITextViewDelegate` in the composite component, in order to change the behaviour, and then forward the calls onto `MultilineTextInput` where necessary.

This is a pretty fragile solution, as it means the new component is fully aware and would depend on the implementation details of `MultilineTextInput` 🤷‍♂️  

If you make the view properties `public` then at least it's clear if a change is made that it would be a breaking change, but what if the `UITextViewDelegate` implementation changes?

For example, for this delegate method:
```swift
    public func textViewShouldBeginEditing(_ textView: UITextView) -> Bool {
        textView.inputAccessoryView = sendButton.isHidden ? FocusToolbar(view: self) : nil
        return true
    }
```

I would have to override this implementation and **not** forward to the existing delegate in order to get the required behaviour.

BentoKit/BentoKit/Components/MultilineTextInput.swift
TheAdamBorek:
```swift
 func pinEdges(to layoutGuide: UILayoutGuide,
                  insets: UIEdgeInsets = .zero,
                  topPriority: UILayoutPriority = .required,
                  bottomPriority: UILayoutPriority = .required,
                  leadingPrioryty: UILayoutPriority = .required,
                  trailingPriority: UILayoutPriority = .required)
```
:)

BentoKit/BentoKit/Components/MultilineTextInput.swift
TheAdamBorek: Ok. However, I don't know if I would assume "send" will resginFirstResponder & call didFinishEditing only looking on the name. I think `button` would also be a good name or `finishEditingButton`. 

Up yo you, that's only a suggestion. I'm not 100% sure it's better ;)

mluisbrown: I'm going to remove the send button anyway and try to use component composition, so this will no longer be an issue.

mluisbrown:
@anders @sergdort @dmcrodrigues I have removed the send button 👍

sergdort:
I'm working on a `Card` component that will allow this for any component ¯\_(ツ)_/¯

mluisbrown:
> I think we need to stop changing BentoKit components based on specifics of our design team.

@sergdort being able to style the content view is a perfectly normal thing to want to do and is not based on the specifics of our design team 🤷‍♂️ 

The content view approach was already added to the `TextInput` component.

sergdort:
> The content view approach was already added to the `TextInput` component.

We did do that but this is simply not scalable. Every time our designers would want to make existing component look like a card (with margins from the edges and corner radius) we would need to make a PR. It also pollutes interface. To some extend this is the same problem we fixed with `Behaviours`, but now we just going to decorate in order to change how it looks

mluisbrown:
@sergdort I removed the `contentView`

PR #116, opened 2019-02-25
Reuse TextField Implementation

gtsiap:
Currently it's not easy to reuse `Components.TextInput`.

This PR introduces `FocusTextField` which contains a `UITextField` and an `AccessoryView`.
The idea is to have compassable views in order to create new components which will be needed.

These changes are source compatible for the consumers of `Components.TextInput`

### DIFF
--- a/BentoKit/BentoKit/Views/FocusTextField.swift
+++ b/BentoKit/BentoKit/Views/FocusTextField.swift
@@ -1,0 +1,132 @@
+import UIKit
+import Bento
+import StyleSheets
+
+public final class FocusTextField: UIView, UITextFieldDelegate {
+    public let accessoryView = AccessoryView()
+    public let textField = UITextField()
+
+    var textWillChange: Optional<(TextChange) -> Bool> = nil
+    var textDidChange: Optional<(String?) -> Void> = nil
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        setupLayout()
+    }
+
+    private func setupLayout() {
+        stack(.horizontal, spacing: 16, distribution: .fill, alignment: .center)(
+            textField, accessoryView
+            )
+            .add(to: self)
+            .pinEdges(to: self)
+
+        textField.addTarget(self, action: #selector(textDidChangeOn(_:)), for: UIControl.Event.editingChanged)
+        textField.delegate = self
+    }
+
+    @available(*, unavailable)
+    public required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    @objc private func textDidChangeOn(_ textField: UITextField) {
+        self.textDidChange?(textField.text)
+    }
+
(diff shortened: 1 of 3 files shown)


sergdort:
I it the same that @mluisbrown did [here](https://github.com/Babylonpartners/Bento/pull/115)?

Also could you please add screenshots

sergdort:
@gtsiap is it a textfieid that highlights its borders when active?

andersio:
Have you explored `UITextField.rightView`?

mluisbrown:
> I it the same that @mluisbrown did [here](https://github.com/Babylonpartners/Bento/pull/115)?

@sergdort this is for TextField, my change it to MultilineTextInput.

TheAdamBorek:
We were talking on #ios-oss slack channel of providing a way to compose multiple components by providing for "layout components" like
`StackViewComponent`(or `Row`/`Column` as Flutter use such naming - naming is not the case here).

Wouldn't that solve the problem you are working on?

gtsiap:
> I it the same that @mluisbrown did [here](https://github.com/Babylonpartners/Bento/pull/115)?
> 
> Also could you please add screenshots

Its not tied to a specific style, that's why I haven't added any screenshots.
You can have a visual of it from here,
`Bento/Snapshots/ReferenceImages_64/BentoKitTests.TextInputSnapshotTests`

gtsiap:
> @gtsiap is it a textfieid that highlights its borders when active?

Its design agnostic. Its not related to a specific stylesheet.

gtsiap:
> Have you explored `UITextField.rightView`?

No, I didn't. I wanted to minimise the changes.

gtsiap:
> We were talking on #ios-oss slack channel of providing a way to compose multiple components by providing for "layout components" like
> `StackViewComponent`(or `Row`/`Column` as Flutter uses such naming but naming is not the case here).
> 
> Wouldn't that solve the problem you are working on?

This discussion is related to this PR.

`StackViewComponent` or something equivalent wouldn't solve my issue because `Components.TextInput` has also a `contentView` which contains the textfield

BentoKit/BentoKit/Views/FocusTextField.swift
```diff
@@ -0,0 +1,123 @@
...
+                else { return "" }
+
+            return current.replacingCharacters(in: range, with: change)
+        }
+    }
+}
+
+extension FocusTextField {
+
+    public enum Accessory {
```
sergdort: Why not use `AccessoryView.Accessory`?

sergdort:
I think we should stop rushing and bounding unnecessary components and adding them into OSS library just because our design team requires it.
as @TheAdamBorek pointed out we can try to introduce `Row/Column` component (in BabylonUI) and see how it goes.

BentoKit/BentoKit/Views/FocusTextField.swift
gtsiap: The initial implementation was restraining  `AccessoryView.Accessory`.
I didn't want to change its behaviour, we can revise this decision.

gtsiap:
I will close this PR for now until we figure out how we will handle compassable components.

PR #117, opened 2019-02-27
Fix button component having multiple buttons

mluisbrown:
Fixes a bug introduced in #99 which means that when the `buttonType` is changed, the old button is not removed from the component view before adding the new button, resulting in two (or more) buttons being added to the component.

### DIFF
--- a/BentoKit/BentoKit/Components/Button.swift
+++ b/BentoKit/BentoKit/Components/Button.swift
@@ -75,6 +75,9 @@ extension Component.Button {
             didSet {
                 guard oldValue != buttonType else { return }
 
+                activityIndicator.removeFromSuperview()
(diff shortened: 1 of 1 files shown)


sergdort:
@mluisbrown  Would it be possible to add unit test for this issue?

mluisbrown:
@sergdort I'm not sure how to reproduce this in a test environment 😕

sergdort:
Render component twice in the same view?

mluisbrown:
> Render component twice in the same view?

I don't think that will work. It only happens when the view is re-used and the style is changed.

sergdort:
Then, probably having two components with different style and render them one by one in the same view

PR #119, opened 2019-03-03
Adds non-frp example

TheAdamBorek:
The current issue with bento is, it's hard to understand for newcomers. I'd like to make it easier to understand how they can write iOS apps with Bento. I added non-frp example written in MVP which I think can make Bento easier to understand for people who don't know ReactiveSwift & ReactiveFeedback.

I also added icons in the tabBar as it's more elegant :D

## How?
I've chosen MVP as I think it's most standard architecture without FRP framework (besides MVC). I stick with having `Renderer` as I think it's good to separate `render` function from other files with small change that ViewController owns the renderer.

## How does it look?
cleanshot 2019-03-03 at 14 09 34


mluisbrown:
Looks great 👍  I personally would call the `SignUpPresenter` the `SignUpViewModel`, but I guess it's fine this way. In this example I can't really see the difference between MVP and MVVM 🤷‍♂️

TheAdamBorek:
@mluisbrown I would say in MVVM you have "Observables" in "MVP" delegates 😂

mluisbrown:
@TheAdamBorek  that's an interesting way to think about it 🤔

sergdort:
Could we please merge this one first 🙈  https://github.com/Babylonpartners/Bento/pull/118

Example/SignUp/SignUpPresetner.swift
```diff
@@ -0,0 +1,138 @@
+import Foundation
+
+protocol SignUpPresenterDelegate: class {
+    func render(_ state: SignUpPresenter.State)
+}
+
+final class SignUpPresenter {
+    weak var delegate: SignUpPresenterDelegate?
```
sergdort: In MVP this usually called `View` 😜

TheAdamBorek:
@sergdort the one who has approval first won the race 🙈

sergdort:
>  the one who has approval first won the race

I think I need to request changes 😆

mluisbrown:
@TheAdamBorek once the conflicts are resolved, this can be merged 👍

PR #121, opened 2019-03-05
Bento in-built pre-sizing layout pass support for multiline text.

andersio:
## Problem Statement
The BentoKit `TitledDescription` component would have a broken layout in the following scenarios, if we removed its manual sizing code:
```
TITLE         |
LINE2         |
--------------|     DETAIL
SUBTITLE      |
LINE2         |
```

Auto Layout apparently doesn't support these scenarios very well, as it produces an insufficient height of the cell that cannot fit all the vertical labels.

The common trick in `layoutSubviews()` alone also isn't applicable, since at that time AL would have already produced the cell size, and whatever `preferredMaxLayoutWidth` we change thereafter does not matter until the next time the UITableView invalidates its layout.

## Solution
One solution is that we would do a two-pass layout — views that have multiline text may opt in by conforming to `PresizingLayoutPassRequiring`, so that Bento would trigger an extra layout pass _before_ the AL cell sizing happens.

Then views containing multiline text labels can then update `preferredMaxLayoutWidth` of labels in `layoutSubviews()`, and being certain that the AL cell sizing happening thereafter would factor in the correct text rect height.

(Apparently UIKit has private APIs for it, and it is used in some unknown cases.)

If no view has specified this, Bento would not trigger a presizing layout pass.

### ~~Removing `HeightCustomizing`.~~
~~Since the primary use case of `HeightCustomizing` would be resolved by the introduction of `RequiringPresizingLayoutPass`, this PR removes also `HeightCustomizing`.~~

~~It may have impact upon scrolling experience in certain places, however, until #80 is deployed. Hence the Do Not Merge flag.~~

### Other
The `TitledDescription` playground has been updated to reproduce the issue & illustrate that the issue is addressed by this new `RequiringPresizingLayoutPass` mechanism.

Before | After
--- | ---
screenshot 2019-03-05 at 10 46 37 | screenshot 2019-03-05 at 10 46 26


mluisbrown:
Just an FYI that when this is merged, and when Babylon is updated to the latest `master` of Bento, there will be breaking changes for internal components that implement `HeightCustomizing`.

andersio:
@mluisbrown Guess I should keep the protocol & just remove the conformance.

Bento/Views/RequiringPreSizingLayoutPass.swift
```diff
@@ -0,0 +1,16 @@
+public protocol RequiringPreSizingLayoutPass {}
```
mluisbrown: I would rename this to `PreSizingLayoutPassRequiring`, which is more in line with the naming convention for this type of protocol, where the infinitive of the verb comes at the end (eg `HeightCustomizing` 😛 )

BentoKit/BentoKit/Components/TitledDescription.swift
```diff
@@ -139,118 +121,6 @@ public extension Component {
...
             self.styleSheet = styleSheet
         }
-
-        private let heightComputer: (CGFloat, UIEdgeInsets) -> CGFloat
-    }
-}
-
-private extension Component.TitledDescription {
-
-    static func heightComputer(
```
mluisbrown: tenor-32591722

mluisbrown:
Did you run the BentoKit snapshot tests, to see if there are any changes?

Bento.xcodeproj/project.pbxproj
```diff
@@ -684,7 +687,7 @@
 				5829D269200FB092001E020D /* AppDelegate.swift in Sources */,
 				04EBE74F222C031B002D38C4 /* Optional+Extension.swift in Sources */,
 				04C9A0842211967300C70E09 /* SignUpViewController.swift in Sources */,
-				61B4091A208523F60063DE25 /* IntroViewController.swift in Sources */,
+				61B4091A208523F60063DE25 /* FoodListViewController.swift in Sources */,
```
Ben-Henshall: Won't this name change also need to happen in the storyboard?

andersio: I didn't change this I believe...

Bento/Views/RequiringPreSizingLayoutPass.swift
```diff
@@ -0,0 +1,16 @@
...
+
+extension UIView {
+    private var needsPresizingLayoutPass: Bool {
+        return self is RequiringPreSizingLayoutPass
+            || subviews.contains(where: { $0.needsPresizingLayoutPass })
```
sergdort: 👍

sergdort:
[Apple are cheaters](https://github.com/JaviSoto/iOS10-Runtime-Headers/blob/master/Frameworks/UIKit.framework/UILabel.h#L185)

Bento/Views/RequiringPreSizingLayoutPass.swift
TheAdamBorek: @andersio can you please add documentation to this protocol? I think it's a protocol people may want to use.
I would describe there our use case

inamiy:
Can you provide the correct playground rather than reproduced one?

andersio:
@inamiy Not sure what do you mean

Bento/Views/RequiringPreSizingLayoutPass.swift
andersio: Done

inamiy:
> Not sure what you mean exactly

Current `TitledDescription.playground` seems to show "Before" screen.

...But wait, I'm also seeing `extension Component.TitledDescription.View: RequiringPreSizingLayoutPass` which I first thought it was not implemented yet... 
Am I missing something?

andersio:
You need to use the updated playground on this branch, while removing `HeightCustomizing` conformance from `TitledDescription` to reveal the issue.

andersio:
The scope of the PR has been reduced to just adding `PresizingLayoutPassRequiring` and updating `TitledDescription`.

It also removes the triggering in `UICollectionViewReusableView` and `UICollectionViewCell`, since Bento does not have sufficient knowledge in these views alone to trigger a presizing layout pass that leads to the expected outcome — it ultimately depends on the layout direction & sizing strategy.

PR #122, opened 2019-03-12
Remove hardcoded size for Toggle image

NSMyself:
The image being used in the Toggle component currently has a hardcoded size of 32x32. This PR aims to fix that by removing that specific size and using intrinsic content size instead.

### DIFF
--- a/BentoKit/BentoKit/Components/Toggle.swift
+++ b/BentoKit/BentoKit/Components/Toggle.swift
@@ -40,19 +40,21 @@ extension Component {
 
             self.heightComputer = { width, inheritedMargins in
                 let verticalMargins = styleSheet.layoutMargins.verticalTotal
-                let imageWidth: CGFloat = image != nil
-                    ? (image?.size.width ?? 0) + Dimensions.horizontalElementsSpacing
-                    : 0.0
+                let imageWidth: CGFloat = image?.size.width ?? 0
+
                 let textBoundWidth = width
                     - max(styleSheet.layoutMargins.left, inheritedMargins.left)
                     - max(styleSheet.layoutMargins.right, inheritedMargins.right)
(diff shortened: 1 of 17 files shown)


mluisbrown:
The change seems not to be working, or at lest the snapshot tests don't reflect this :/

NSMyself:
@mluisbrown i've added two new tests for this new behaviour, please have a look when possible

BentoKit/BentoKit/Components/Toggle.swift
```diff
@@ -40,19 +40,21 @@ extension Component {
 
             self.heightComputer = { width, inheritedMargins in
                 let verticalMargins = styleSheet.layoutMargins.verticalTotal
-                let imageWidth: CGFloat = image != nil
-                    ? Dimensions.imageViewWidth + Dimensions.horizontalElementsSpacing
-                    : 0.0
+                let imageWidth: CGFloat = image?.size.width ?? 0
```
andersio: Between the image and the label, the stack view spacing applies when the UIImageView is not hidden.

NSMyself: Thanks for pointing that out, wasn't aware @andersio. Will fix that asap

mluisbrown:
Previously the test `test_toggle_with_image` used a 32x32 image (because it was hardcoded). Since the test uses `.image(fromColor: .blue)`, which defaults to a size of 1x1 IMO the test should be changed to use a 32x32 image (`.image(fromColor: .blue, size: CGSize(width: 32, height: 32))`, as a 1px image is almost invisible, and it would mean the snapshots for this test wouldn't need to change.

NSMyself:
Fair point 👍 Changing

PR #127, opened 2019-03-13
Remove `<>` operator

### LINKED ISSUE #126: Remove <> operator
In Bento 0.4.0 we want to remove `<>` operator as we think usage of Node.init is verbose enough and it doesn't need an operator to simplify it.

Moreover, `<>` is a known operator in the functional programming world. We don't want to Bento stays in the conflict with it.

inamiy:
**Note: Breaking change.**

Resolves #126 .

### DIFF
--- a/Example/Basic/ViewController.swift
+++ b/Example/Basic/ViewController.swift
@@ -126,17 +126,23 @@ class ViewController: UIViewController {
     private func renderIconText() -> Node<RowId> {
         switch state {
         case .airplaneMode:
-            return Node(id: RowId.note, component: IconTextComponent(image: #imageLiteral(resourceName: "wifi"), title: "WIFI Off")
-                .deletable(deleteActionText: "Delete", didDelete: {
-                    print("Delete")
-                })
-                .on(willDisplayItem: nil, didEndDisplayingItem: nil))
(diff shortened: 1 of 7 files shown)


Example/Basic/ViewController.swift
```diff
@@ -120,23 +120,23 @@ class ViewController: UIViewController {
...
                                         })
-        return RowId.toggle <> component
+        return Node(id: RowId.toggle, component: component)
     }
 
     private func renderIconText() -> Node<RowId> {
         switch state {
         case .airplaneMode:
-            return RowId.note <> IconTextComponent(image: #imageLiteral(resourceName: "wifi"), title: "WIFI Off")
+            return Node(id: RowId.note, component: IconTextComponent(image: #imageLiteral(resourceName: "wifi"), title: "WIFI Off")
```
TheAdamBorek: Can I ask you to change the indentation to keep syle from the main project?

```
Node(
    id: .securityAnswer,
    component: IconTextComponent(
        <params>
    )
)
```

What do you think? 
That would apply to all changes here

inamiy: Done in f70f003 👍

PR #128, opened 2019-03-13
Remove `If<T>` & `Some<T, U>`

inamiy:
**Note: Breaking change.**

This PR removes `If<T>` and `Some<T, U>` types as discussed in https://github.com/Babylonpartners/Bento/pull/123#issuecomment-471929681 by using plain `Optional` instead.

### DIFF
--- a/Bento/Bento/If.swift
+++ b/Bento/Bento/If.swift
@@ -1,5 +1,5 @@
 extension Optional {
-    public static func iff(_ condition: @autoclosure @escaping () -> Bool, _ generator: @autoclosure @escaping () -> Wrapped) -> Wrapped? {
+    public static func iff(_ condition: @autoclosure () -> Bool, _ generator: @autoclosure () -> Wrapped) -> Wrapped? {
         if condition() {
             return generator()
         } else {
@@ -7,7 +7,7 @@ extension Optional {
         }
     }
 
-    public static func iff(_ condition: @autoclosure @escaping () -> Bool, _ generator: @escaping () -> Wrapped) -> Wrapped? {
+    public static func iff(_ condition: @autoclosure () -> Bool, _ generator: () -> Wrapped) -> Wrapped? {
         if condition() {
             return generator()
         } else {
@@ -15,7 +15,7 @@ extension Optional {
         }
     }
 
-    public static func iff(_ condition: @escaping () -> Bool, _ generator: @escaping () -> Wrapped) -> Wrapped? {
+    public static func iff(_ condition: () -> Bool, _ generator: () -> Wrapped) -> Wrapped? {
         if condition() {
             return generator()
         } else {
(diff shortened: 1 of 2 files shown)


Bento/Bento/If.swift
```diff
@@ -1,99 +1,69 @@
...
-    let condition: () -> Bool
-    let generator: () -> T
-
-    public static func iff(_ condition: @autoclosure @escaping () -> Bool, _ generator: @autoclosure @escaping () -> T) -> If<T> {
-        return If(condition: condition, generator: generator)
-    }
-
-    public static func iff(_ condition: @autoclosure @escaping () -> Bool, _ generator: @escaping () -> T) -> If<T> {
-        return If(condition: condition, generator: generator)
+extension Optional {
```
TheAdamBorek: shouldn't we add type constraints here? `where Wrapped == <BentoTypes>` ?

BentoTests/ConcatenationTests.swift
```diff
@@ -70,7 +70,7 @@ class ConcatenationTests: XCTestCase {
         let optional: String? = "something"
 
         let result: Section<TestSectionId, TestRowId> = section
-            |---? .some(optional) { _ in
+            |---? optional.map { _ in
                 node
             }
```
TheAdamBorek: Do we have tests for `.iff` ?

If not can you please add one in this file?

TheAdamBorek:
👍 
one question, one request for a test - that's all

BentoTests/ConcatenationTests.swift
inamiy: `|-? .iff` tests go to `IfTests.swift`, which are unchanged in this PR.

Bento/Bento/If.swift
inamiy: `Section` / `[Section]` / `Node` / `[Node]` will be `Wrapped` in Bento, and the implementation itself is very generic, so I think we should keep this generic form.

TheAdamBorek: ahhhh, nevermind. its not a public extension
👍

ilyapuchka: the methods are public though, so it will be accessible for any optional, but as @inamiy said they are generic.

Bento/Bento/If.swift
```diff
@@ -1,99 +1,69 @@
...
+    public static func iff(_ condition: @autoclosure @escaping () -> Bool, _ generator: @autoclosure @escaping () -> Wrapped) -> Wrapped? {
```
ilyapuchka: @inamiy I don't think `@escaping` is really needed, can we remove it? Because of this we have to use `self` in each `.iff`/`.some` call =/

inamiy: Good catch 👍  Fixed in 437bd22 .

sergdort:
It's not a breaking change change, is it?

inamiy:
@sergdort 

```diff
-    |-? .some(optional) { _ in
+    |-? optional.map { _ in
```

`Some.some` is gone :)

sergdort:
@inamiy Could you please make a PR to the main repo which removes it?

PR #138, opened 2019-04-01
Expose contentView spacing via Toggle builder

NSMyself:
In the toggle component we need to be able to specify the spacing between the image (left side) and the actual title because TD has a different value and, as such, the labels of these components when placed next to each other are misalign. 

By allowing the injection of this spacing amount using Toggle's initializer (and a default value of 9, which is currently what is being used) we can effectively fix this problem.

### DIFF
--- a/BentoKit/BentoKit/Components/Toggle.swift
+++ b/BentoKit/BentoKit/Components/Toggle.swift
@@ -15,6 +15,7 @@ extension Component {
             isOn: Bool = false,
             isEnabled: Bool = true,
             isRefreshing: Bool = false,
+            horizontalSpacing: CGFloat = 9,
             animateValueChange: Bool = false,
             styleSheet: StyleSheet,
             didChangeValue: ((Bool) -> Void)? = nil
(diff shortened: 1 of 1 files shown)


BentoKit/BentoKit/Components/Toggle.swift
```diff
@@ -137,14 +137,17 @@ extension Component.Toggle {
 extension Component.Toggle {
     public final class StyleSheet: BaseViewStyleSheet<View> {
         public let text: LabelStyleSheet
+        public let spacing: CGFloat
 
-        public init(text: LabelStyleSheet) {
+        public init(text: LabelStyleSheet, spacing: CGFloat = 9) {
```
RuiAAPeres: Why 9?

NSMyself: 9 is the [default value currently being used](https://github.com/Babylonpartners/Bento/blob/c81f6ccb5df2cc61d80df3bc62728ea3c32b4c1f/BentoKit/BentoKit/Components/Toggle.swift#L105).

RuiAAPeres: But the `9` is still there, it's not being read from `spacing`.

NSMyself: We're overriding this value later, essentially, when we run:

RuiAAPeres: I see. 👍

NSMyself: If you want i can look into alternatives to prevent this, of course

PR #141, opened 2019-04-04
Introduce `AdapterStore` and integrate it with the TableView adapters.

andersio:
Refer to https://github.com/Babylonpartners/Bento/issues/80 for the implementation plan.

----

`AdapterStore` is introduced to support size caching in both UITableView and UICollectionView adapters.

It is responsible of holding all sections of items, and lazily requests size via Auto Layout at a particular index path when it receives such request. Moreover, whenever Bento renders a new box, it is also responsible of updating its internal data structure correctly in regards to the changeset.

It also tracks attributes like container view bound size and the layout margins, so that it can trigger Auto Layout calls with correct parameters, and most importantly, invalidate all cache items when these attributes change.

At the moment, the adapter is integrated with `TableViewAdapterBase`, but with size caching **disabled**. It is expected that ONLY the `SizeCachingTableView` (to be introduced) would have size caching turned on.

### DIFF
--- a/Bento/Adapters/AdapterStore.swift
+++ b/Bento/Adapters/AdapterStore.swift
@@ -103,7 +103,7 @@ struct AdapterStore<SectionID: Hashable, ItemID: Hashable> {
             newElement: SectionInfo(),
             whenInserted: { info, index in
                 info.itemSizes = Array(repeating: .unknown, count: sections[index].items.count)
-        }
+            }
         )
 
         // Apply changeset to the old section info for all mutated sections.
--- a/Bento/Adapters/AdapterStore.swift
+++ b/Bento/Adapters/AdapterStore.swift
@@ -1,0 +1,207 @@
+import FlexibleDiff
+
+/// The adapter store which carries the component tree, and also cached sizing information if enabled.
+///
+/// The adapter computes sizing information lazily upon request, but it adjusts spaces and performs cache invalidation
+/// every time a diff is applied.
+///
+/// - important: While the use-as-you-go Bento uses the adapter, the size caching capability should only be enabled in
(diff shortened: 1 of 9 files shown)


Bento/Adapters/AdapterStore.swift
```diff
@@ -0,0 +1,207 @@
...
+
+/// The adapter store which carries the component tree, and also cached sizing information if enabled.
+///
+/// The adapter computes sizing information lazily upon request, but it adjusts spaces and performs cache invalidation
+/// every time a diff is applied.
+///
+/// - important: While the use-as-you-go Bento uses the adapter, the size caching capability should only be enabled in
+///              `SizeCachingTableView` and `SizeCachingCollectionView`, since there are specific messages that Bento
+///              need to intercept for size caching to function as expected.
+struct AdapterStore<SectionID: Hashable, ItemID: Hashable> {
```
sergdort: Why making it a struct? It seems to me that it behaves more like a ref type, no?

andersio: It doesn't need identity. After all it is just operations upon a bunch of data.

Bento/Adapters/AdapterStore.swift
```diff
@@ -0,0 +1,207 @@
...
+            resetCachedInfo()
+            return
+        }
+
+        info.applyIgnoringMutation(
+            changeset.sections,
+            newElement: SectionInfo(),
+            whenInserted: { info, index in
+                info.itemSizes = Array(repeating: .unknown, count: sections[index].items.count)
+        }
```
mluisbrown:
```suggestion
            }
```

PR #142, opened 2019-04-08
Improve documentation for `BentoReusableView`

inamiy:
This PR improves documentation for `BentoReusableView` as follows:

- Distinguishes terminology between `contentView` and `containedView`
- Describes the relationship with `ViewLifecycleAware` and `ComponentLifecycleAware`
- Renames internal method for readability

### DIFF
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -1,4 +1,4 @@
-/// Existential type for reusable views (cell, header, footer) in `UITableView`, `UICollectionView`, etc.
+/// Protocol for reusable views (cell, header, footer) in `UITableView`, `UICollectionView`, etc.
 protocol BentoReusableView: AnyObject {
     /// The view to which `containedView` should be added.
     /// - Note: This is normally system-provided as `self`'s subview.
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -14,7 +14,7 @@ protocol BentoReusableView: AnyObject {
 }
 
 extension BentoReusableView {
-    /// Updates `component` to either replace `containedView` with a new one or reuse it.
+    /// Uses `component` to either replace `containedView` with a new one or reuse it.
     func bind(_ component: AnyRenderable?) {
         self.component = component
         if let component = component {
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -1,6 +1,6 @@
 /// Existential type for reusable views (cell, header, footer) in `UITableView`, `UICollectionView`, etc.
 protocol BentoReusableView: AnyObject {
-    /// The main view to which you add your reusable view’s custom content.
+    /// The view to which `containedView` should be added.
     /// - Note: This is normally system-provided as `self`'s subview.
     var contentView: UIView { get }
 
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -5,7 +5,7 @@ protocol BentoReusableView: AnyObject {
     var contentView: UIView { get }
 
     /// `component`-generated view that will be added to `contentView`.
-    /// - Note: Subtype of this instance can conforming to `ViewLifecycleAware` to allow display handling.
+    /// - Note: Subtype of this instance can conform to `ViewLifecycleAware` to allow display handling.
     var containedView: UIView? { get set }
 
     /// A renderer which generates `containedView`.
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -1,10 +1,20 @@
+/// Existential type for reusable views (cell, header, footer) in `UITableView`, `UICollectionView`, etc.
 protocol BentoReusableView: AnyObject {
-    var containedView: UIView? { get set }
+    /// The main view to which you add your reusable view’s custom content.
+    /// - Note: This is normally system-provided as `self`'s subview.
     var contentView: UIView { get }
+
+    /// `component`-generated view that will be added to `contentView`.
+    /// - Note: Subtype of this instance can conforming to `ViewLifecycleAware` to allow display handling.
+    var containedView: UIView? { get set }
+
+    /// A renderer which generates `containedView`.
+    /// - Note: Underlying type of this instance can conform to `ComponentLifecycleAware` to allow display handling.
     var component: AnyRenderable? { get set }
 }
 
 extension BentoReusableView {
+    /// Updates `component` to either replace `containedView` with a new one or reuse it.
     func bind(_ component: AnyRenderable?) {
         self.component = component
         if let component = component {
@@ -39,7 +49,8 @@ extension BentoReusableView {
 }
 
 extension BentoReusableView where Self: UIView {
-    func containerViewDidChange(from old: UIView?, to new: UIView?) {
+    /// - Note: Call this method when old `containedView` is replaced with new one.
+    func containedViewDidChange(from old: UIView?, to new: UIView?) {
         func add(_ view: UIView) {
             contentView.addSubview(view)
             view.pinToEdges(of: contentView)
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -9,7 +9,7 @@ protocol BentoReusableView: AnyObject {
     var containedView: UIView? { get set }
 
     /// A renderer which generates `containedView`.
-    /// - Note: Underlying type of this instance can conform to `ComponentLifecycleAware` to allow display handling.
+    /// - note: Underlying type of this instance can conform to `ComponentLifecycleAware` to allow display handling.
     var component: AnyRenderable? { get set }
 }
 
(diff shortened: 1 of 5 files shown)


Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
+/// Existential type for reusable views (cell, header, footer) in `UITableView`, `UICollectionView`, etc.
 protocol BentoReusableView: AnyObject {
-    var containedView: UIView? { get set }
+    /// The main view to which you add your reusable view’s custom content.
+    /// - Note: This is normally system-provided as `self`'s subview.
     var contentView: UIView { get }
+
+    /// `component`-generated view that will be added to `contentView`.
+    /// - Note: Subtype of this instance can conforming to `ViewLifecycleAware` to allow display handling.
```
mluisbrown:
```suggestion
    /// - Note: Subtype of this instance can conform to `ViewLifecycleAware` to allow display handling.
```

mluisbrown (approved):

Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
+/// Existential type for reusable views (cell, header, footer) in `UITableView`, `UICollectionView`, etc.
```
andersio: A protocol is not an existential.

Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
...
 protocol BentoReusableView: AnyObject {
-    var containedView: UIView? { get set }
+    /// The main view to which you add your reusable view’s custom content.
```
andersio:
```suggestion
    /// The view to which `containedView` should be added.
```

Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
...
+    /// - Note: This is normally system-provided as `self`'s subview.
```
andersio, line 4: This should be a precondition.

Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
...
     var contentView: UIView { get }
+
+    /// `component`-generated view that will be added to `contentView`.
+    /// - Note: Subtype of this instance can conform to `ViewLifecycleAware` to allow display handling.
+    var containedView: UIView? { get set }
+
+    /// A renderer which generates `containedView`.
```
andersio, line 13: An instance of component does not generate a view. They only bind with one.

Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
+/// Existential type for reusable views (cell, header, footer) in `UITableView`, `UICollectionView`, etc.
 protocol BentoReusableView: AnyObject {
-    var containedView: UIView? { get set }
+    /// The main view to which you add your reusable view’s custom content.
+    /// - Note: This is normally system-provided as `self`'s subview.
     var contentView: UIView { get }
+
+    /// `component`-generated view that will be added to `contentView`.
```
andersio:
```suggestion
    /// The component root view that is currently installed in `contentView`. If the view is not compatible with the component to be bound, a new compatible view should be instantiated to replace it.
```

Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
...
+    /// - Note: Subtype of this instance can conform to `ViewLifecycleAware` to allow display handling.
+    var containedView: UIView? { get set }
+
+    /// A renderer which generates `containedView`.
+    /// - Note: Underlying type of this instance can conform to `ComponentLifecycleAware` to allow display handling.
```
andersio:
```suggestion
    /// - note: Underlying type of this instance can conform to `ComponentLifecycleAware` to allow display handling.
```

Bento/Views/BentoReusableView.swift
```diff
@@ -1,10 +1,20 @@
...
     var component: AnyRenderable? { get set }
 }
 
 extension BentoReusableView {
+    /// Updates `component` to either replace `containedView` with a new one or reuse it.
```
andersio: This does not accurately describe the relationship. We don't update the component — the component is opaque to Bento logic. We inspect whether the view is compatible with the component, create a new view as appropriate, and finally bind the component with the view.

Bento/Views/BentoReusableView.swift
```diff
@@ -39,7 +49,8 @@ extension BentoReusableView {
 }
 
 extension BentoReusableView where Self: UIView {
-    func containerViewDidChange(from old: UIView?, to new: UIView?) {
+    /// - Note: Call this method when old `containedView` is replaced with new one.
```
andersio:
```suggestion
    /// - important: This should be invoked whenever `containedView` is changed.
```

andersio (changes requested):

Bento/Views/BentoReusableView.swift
inamiy: Can you provide more information about why you think so?

P.S. `protocol  BentoReusableView` is currently _used_ an existential only.

Bento/Views/BentoReusableView.swift
inamiy: Proposed edition provides less information about `contentView`.
If we see `containedView`'s comment, it already provides the hierarchy, so we don't need to mention twice.

(BTW, this comment is taken from UIKit's original comment, which I think is reasonable)

Bento/Views/BentoReusableView.swift
inamiy: I think we don't need to force this as precondition since this could be useful in other view-reusing library which might not have the same variable name.

Bento/Views/BentoReusableView.swift
inamiy: `AnyRenderable` owns `viewType` which generates `containedView` _indirectly_ .

Bento/Views/BentoReusableView.swift
inamiy: Can you provide more info about `If the view is not compatible...` ?
I think the scenario example needs to be also documented.

Bento/Views/BentoReusableView.swift
inamiy: f494d65 Fixed the comment with `Uses` rather than `Updates`.

Bento/Views/BentoReusableView.swift
sergdort: `contentView` is not always `self` e.g `UITableViewCell` and `UITableViewHeaderFooterView`

sergdort (approved):

Bento/Views/BentoReusableView.swift
inamiy: @sergdort  Here I wrote `contentView` as `self`'s subview, not `self`.

Bento/Views/BentoReusableView.swift
andersio: It is purely terminology. How a protocol is used, generics or existentials, does not change what the protocol is, which is a static-time contract with semantics.

Bento/Views/BentoReusableView.swift
andersio: `AnyRenderable` does not "own" `viewType`. We do query the runtime type through an instance property, but it doesn't mean saying a `Renderable` _instance_ generating the view is semantically right. It is primarily because Swift does not support [generalized existentials and opening them up](https://github.com/apple/swift/blob/master/docs/GenericsManifesto.md#opening-existentials) yet, and therefore we need it to be done this way for querying the associated types.

Bento/Views/BentoReusableView.swift
andersio: If the cell is previously bound with Component Apple, and now it is asked to be bound with Component Banana, the current `containedView` would not be compatible with that of Component Banana, and therefore a view needs to be instantiated for Component Banana.

Bento/Views/BentoReusableView.swift
andersio: > Proposed edition provides less information about contentView.

In the context of `BentoReusableView`, the `contentView` requirement for Bento is as simple as described — "custom content" is `containedView` and would stay as such based on today's knowledge.

sergdort:
Any update on this one?

Bento/Views/BentoReusableView.swift
inamiy:
```suggestion
    /// The component root view that is currently added in `contentView`.
    /// If the existing view is no longer compatible with the newly set `component`,
    /// a new compatible view should be instantiated to replace it.
```

Slightly modified https://github.com/Babylonpartners/Bento/pull/142#discussion_r273426139 for simplicity.

Bento/Views/BentoReusableView.swift
inamiy:
```swift
public struct AnyRenderable: Renderable {
    /// The runtime view type of the wrapped `Renderable`.
    public var viewType: NativeView.Type {
        return base.viewType
    }
```

It owns `base` though. (I don't care about owning directly or indirectly)
And `AnyRenderable.viewType.generate()` eventually means "A renderer which generates `containedView`".

Bento/Views/BentoReusableView.swift
inamiy: 0aeae60 Updated to not use jargon.

PR #145, opened 2019-04-09
Upgrade BoxViewController to use the size caching BoxViewController.

andersio:
As per the implementation plan for size caching in #80, this is the third step.

It makes `BoxViewController` use the newly introduced `SizeCachingTableView` for all of its table views, so that everything gets size cached by default.

`HeightCustomizing` is still applicable since the adapter used by `BoxViewController` falls back to the default adapter behavior only when the `HeightCustomizing` conformance is absence. In any case, it is bound to be removed in the next PR, which is going to be the rebased #121.

P.S. We have to duplicate `BentoTableView` to `BoxSizeCachingTableView`, since otherwise all Forms screens would be broken due to instantiation requirement changes.


andersio:
Tests are failing due to #147.

TheAdamBorek:
@andersio is the PR basically only about extending a `BoxSizeCachingTableView` from `SizeCachingTableView`? The rest is just copy-paste?

andersio:
@TheAdamBorek Yes except https://github.com/Babylonpartners/Bento/pull/145/commits/8f06ec240d8d14bc212c4b729db7f84d4014f375 which makes an existing component to work correctly with AL.

sergdort:
@andersio Is it still blocked?

PR #146, opened 2019-04-09
Reusability hint to improve reuse performance when composite components are present.

andersio:
The purpose of this change is pretty much summed up [in the documentation for `Renderable.makeReusabilityHint(using:)`](https://github.com/Babylonpartners/Bento/blob/f34490fcd280719897484599c5f30b458ea78199/Bento/Renderable/Renderable.swift).

The design is inspired by how `Hashable` handles nested hashing.

Do also note that implementing `makeReusabilityHint(using:)` does not imply that a composite component no longer needs to handle type mismatch in view hierarchy.


Bento/Renderable/Renderable.swift
```diff
@@ -1,14 +1,72 @@
...
+    /// - note: This is an optional requirement intended for composite components that contain children components
+    ///         or dynamic view hierarchies.
+    ///
+    /// - parameters:
+    ///   - combiner: The combiner to concatenate all relevant information that affects reusability.
+    func makeReusabilityHint(using combiner: inout ReusabilityHintCombiner)
 }
 
 public extension Renderable {
+    func makeReusabilityHint(using combiner: inout ReusabilityHintCombiner) {}
```
sergdort: Does it scan it through the hierarchy? e.g `C(children: [C(children: [A()], B()])` vs C(children: [C(children: [B()], A()])?

Bento/Renderable/ReusabilityHintCombiner.swift
```diff
@@ -0,0 +1,27 @@
...
+    internal init<R: Renderable>(root: R) {
+        types = [root.componentType]
+    }
+
+    public mutating func combine<R: Renderable>(_ component: R) {
+        types.append(component.componentType)
+        component.makeReusabilityHint(using: &self)
+    }
+
+    __consuming func generate() -> String {
```
sergdort: `__consuming` what does it do?

sergdort (approved):

Bento/Renderable/Renderable.swift
andersio: It doesn’t do so automatically. When you say `combiner.combine(child)`, A’s part of the call it would eventually invoke `makeReusabilityHint(using:)` on the child. So as long as you combine every child using the combiner, and that all children implement this properly, the whole hierarchy would be walked over.

This can be made less crumblesome if we have a protocol for composite component e.g. `CompositeRenderable` which is also useful for auto propagation of messages like `*LifecycleAware`. But this can be done after this PR which lays the cornerstone.

Bento/Views/BentoReusableView.swift
```diff
@@ -6,11 +6,15 @@ protocol BentoReusableView: AnyObject {
...
+        let oldComponent = self.component
         self.component = component
+
         if let component = component {
             let renderingView: UIView
 
-            if let view = containedView, type(of: view) == component.viewType {
+            if let oldComponent = oldComponent,
+               let view = containedView,
+               oldComponent.reusabilityHintCombiner.isCompatible(with: component.reusabilityHintCombiner) {
```
sergdort: This should not usually happen, should it?

andersio: All visible items hit this path when the state changes.

andersio:
I actually realise when answering Sergey’s questions about walking the hierarchy, that we need to generate the reuse identifier with not only `,` but brackets that indicate containment. Otherwise, `A([A([B(), C()])])` would resolve to the same symbol as `A([A(), B(), C()])` which has a rather different view hierarchy.

That’s also reiterate the value of closing off custom `reuseIdentifier`, because who knows what surprises you could have when you have components from different authors. 👻

Bento/Views/BentoReusableView.swift
sergdort: I mean `oldComponent.reusabilityHintCombiner.isCompatible(with: component.reusabilityHintCombiner)` to be `false`. As it'f false it means that we are recreating the view, right?

sergdort (approved):

Bento/Views/BentoReusableView.swift
andersio: yeh

andersio:
Need to think about how this should integrate with #162.

Issue #144, opened 2019-04-08
Why the separation between Bento / BentoKit / StyleSheets?


mluisbrown:
## Why the separation between Bento / BentoKit / StyleSheets?

Is anyone realistically going to use `Bento` on its own? I don’t really think so, because it has no ready to use components, so only a masochist would go down that route. However, `BentoKit`, which gives you some components, obliges you to integrate ReactiveSwift/ReactiveCocoa (RAS/RAC) into your project, even if you don’t want to use it.

None of the `Component`s in `BentoKit` require RAS/RAC except `TitledDescription` 🤦‍♂️, which only uses it for the `image` property.

In fact, the *only* things in `BentoKit` that require RAS/RAC are: `BoxViewModel`, `BoxViewController` and `ImageOrLabelView` (and therefore `TitledDescription`). For `ImageOrLabelView` it’s only a `Reactive` extension which could easily be moved elsewhere.

`StyleSheets` is also required to use the `Component`s so TBH I don’t see why it needs to be in a separate framework (it’s not particularly useful on its own)

If we want `Bento` to be developer friendly ideally we should:

- move all the `Component`s from `BentoKit` into `Bento` (except `TitledDescription`) 
- move what remains of `BentoKit` such as all the RAS/RAC things like `BoxViewModel`, `BoxViewController` and all that they are part of (`Screen` etc) into a new library into a separate repo.

That way in `Bento` we would be left with a nice declarative, component based, tableview / collection view library (`Bento` + components + `StyleSheets`) that anyone could use without being wedded to RAS/RAC or the `Screen` / `BoxViewController` / `BoxViewModel` architecture. 

`BentoKit` would become its own thing that depends on `Bento`.

TheAdamBorek:
I like the idea of:
> That way in Bento we would be left with a nice declarative, component based, tableview / collection view library (Bento + components + StyleSheets) that anyone could use without being wedded to RAS/RAC or the Screen / BoxViewController / BoxViewModel architecture.

The only thing I keep thinking is if we should do this now if we're planning to introduce atomic components. Maybe it's better to hold until atomic components are part of bento and don't merge BentoKit?

mluisbrown:
@TheAdamBorek I think we should do it now. If we keep waiting for "the next big thing" we will never do anything.

TheAdamBorek:
yes, you are right. We don't release too often :/

TheAdamBorek:
[...]

PR #148, opened 2019-04-16
Move all Components and StyleSheets to Bento

mluisbrown:
As discussed in #144 we are going to move everything that doesn't require ReactiveSwift to Bento.

This will include all Components (and their views) from BentoKit, and all StyleSheets. StyleSheets will cease to exist as a separate framework. BentoKit will continue on with the `Screen`, `BoxRenderer` etc components that depend on ReactiveSwift.

This PR also creates the `DetailedDescription` component in `Bento` which is exactly `TitledDescription` but without a reactive image property. The SignUp example uses this component now instead of `TitledDescription` and no longer depends on `BentoKit`

- [x] Move all Components to Bento
- [x] Move component playground to Bento
- [x] Move all StyleSheets to Bento
- [x] Remove all ReactiveFeedback examples
- [x] Update README


Bento/Components/DetailedDescription.swift
```diff
@@ -0,0 +1,439 @@
...
+    /// (1) [Optional] an image or a text placeholder, at most 100 points in width,
+    ///     with an optional badge icon whose position is customisable;
+    /// (2) [Optional] a vertical stack of zero or more independent text labels;
+    /// (3) [Optional] a single-line detail label; and
+    /// (4) [Optional] an accessory.
+    ///
+    /// While `DetailedDescription` supports asynchronous loading for its image
+    /// view, you are obligated to ensure a consistent fixed size across all
+    /// changes.
+    public final class DetailedDescription: AutoRenderable, Focusable {
```
TheAdamBorek: this is `TitledDescription` but without Property?

mluisbrown: Yes, it's exactly the same, except that the image property is not a reactive `Property`.

PR #149, opened 2019-04-16
Propagate `ViewLifecycleAware` calls to the entire view hierarchy.

andersio:
... so that all views can reliably get notified of the visibility of the component.

### DIFF
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -27,14 +27,20 @@ extension BentoReusableView {
         component?
             .cast(to: ComponentLifecycleAware.self)?
             .willDisplayItem()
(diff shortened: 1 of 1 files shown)


Bento/Views/BentoReusableView.swift
```diff
@@ -66,3 +72,13 @@ extension BentoReusableView where Self: UIView {
         }
     }
 }
+
+fileprivate extension UIView {
+    func enumerateAllViews(_ action: (UIView) -> Void) {
```
mluisbrown: `enumerateAllViews` is not very clear about what this does 😕  Perhaps `forViewAndAllSubviews` would make it clearer what is actually happening?

PR #157, opened 2019-04-18
Improve performance of estimated height queries.

andersio:
`SizeCachingTableView` currently answers all estimated height queries with the final size of the component. This causes significant performance toll when initialising screens with large amount of items.

This PR changes it to respond with a quick estimation, which can be

* the last known (but invalidated) height of the ID path; or in its absence
* the estimated height as defined by the `SizeCachingTableView`.

### DIFF
--- a/Bento/Adapters/TableViewAdapter.swift
+++ b/Bento/Adapters/TableViewAdapter.swift
@@ -75,7 +75,7 @@ open class TableViewAdapterBase<SectionID: Hashable, ItemID: Hashable>
 
     @objc open func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
         switch store.size(for: .header, inSection: section) {
-        case .cachingDisabled:
+        case .noCachedResult:
             return tableView.sectionHeaderHeight
         case .doesNotExist:
             return .leastNonzeroMagnitude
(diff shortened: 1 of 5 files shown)


Bento/Adapters/TableViewAdapter.swift
```diff
@@ -125,7 +125,8 @@ open class TableViewAdapterBase<SectionID: Hashable, ItemID: Hashable>
 
     @objc(tableView:estimatedHeightForRowAtIndexPath:)
     open func tableView(_ tableView: UITableView, estimatedHeightForRowAt indexPath: IndexPath) -> CGFloat {
-        return store.size(forItemAt: indexPath).map { $0.height + tableView.separatorHeight }
+        return store.size(forItemAt: indexPath, allowEstimation: true)
+            .map { $0.height + tableView.separatorHeight }
```
sergdort: So is it now lazily computes the height?

andersio: It doesn't compute the height when the `UITableView` asks for the estimated height.

In presence of estimated height, `UITableView` is supposed to call `heightForRowAtIndexPath:` only when the cell is about to enter the prefetching & visible sliding window.

mluisbrown (approved):

dmcrodrigues (approved):

PR #158, opened 2019-04-24
Custom Input with Highlighting Gesture behavior.

andersio:
1. Removed the custom input parameter from `DetailedDescription`.
2. The any-box-as-input-view feature is now available as the `customInput` behavior.
3. Updated the SignUp example.

### DIFF
--- a/Bento/Renderable/CustomInputComponent.swift
+++ b/Bento/Renderable/CustomInputComponent.swift
@@ -4,28 +4,25 @@ struct CustomInputComponent<Base: Renderable>: Renderable, Focusable {
     let customInput: CustomInput
     let focusEligibility: FocusEligibility
     let highlightColor: UIColor?
-    let backgroundColor: UIColor
     let base: Base
 
     init(
         source: Base,
         customInput: CustomInput,
         contentStatus: FocusEligibility.ContentStatus,
-        highlightColor: UIColor?,
-        backgroundColor: UIColor
+        highlightColor: UIColor?
     ) {
         self.customInput = customInput
         self.highlightColor = highlightColor
-        self.backgroundColor = backgroundColor
         self.base = source
         self.focusEligibility = .eligible(contentStatus)
     }
 
     func render(in view: ComponentView) {
         view.inputNodes = customInput
-        view.backgroundColor = backgroundColor
         view.highlightingGesture.didTap = .manual
         view.highlightingGesture.highlightColor = highlightColor
+        view.highlightingGesture.stylingView = view.containedView
 
         base.render(in: view.containedView)
     }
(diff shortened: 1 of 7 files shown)


Bento/Renderable/CustomInputComponent.swift
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        private func neighboringFocusEligibilityDidChange() {
+            focusToolbar?.updateFocusEligibility(with: self)
+            reloadInputViews()
+        }
+
+        @objc func keyboardDidDisappear() {
+            if isFirstResponder { _ = resignFirstResponder() }
+            customInputView = nil
```
sergdort: Why do we need to make it `nil`?

Bento/Renderable/CustomInputComponent.swift
```diff
@@ -0,0 +1,119 @@
...
+            focusToolbar = nil
```
sergdort: Dito

Bento/Renderable/Renderable.swift
```diff
@@ -33,4 +33,19 @@ public extension Renderable {
             didEndDisplayingItem: didEndDisplayingItem
         ).asAnyRenderable()
     }
+
+    func customInput(
+        _ input: CustomInput,
+        contentStatus: FocusEligibility.ContentStatus = .empty,
+        highlightColor: UIColor? = UIColor(red: 239/255.0, green: 239/255.0, blue: 244/255.0, alpha: 1),
+        backgroundColor: UIColor = .white
```
sergdort: Why do we need a background color? We are just wrapping it into a decoration view which can have a clear, right?

Bento/Renderable/CustomInputComponent.swift
andersio: Release unused view.

Bento/Renderable/Renderable.swift
andersio: Forgot that I invented `HighlightingGesture.stylingView`.

Bento/Renderable/CustomInputComponent.swift
andersio: This is copy pasta from DetailedDescription anyway. :)

sergdort: I'm also not sure why do we need to observe key board notifications

andersio: According to my vague memory, these need to stay until the system input view proxy thingy is done with the view.

andersio: Argh. The toolbar would stick around after interactive swipe-to-dismiss-keyboard gesture. This is to make the view to resign from being the first responder.

PR #162, opened 2019-05-04
[CNSMR-1590] New `willMount` and `willUnmount` hooks to support atomic component DSLs.

andersio:
https://babylonpartners.atlassian.net/browse/CNSMR-1590

Two new hooks so we won't need to create wrapping views for certain classes of additive behaviours, and won't need to rely on ObjC associated objects. 

### Styling
The branch includes a pilot which applies a `StyleSheet` to the view using `didMount` and reverts it in `willUnmount`.

```swift
let styleSheet = StyleSheet()
    .setting(\.backgroundColor, .blue)

let styledComponent: AnyRenderable = component.styling(styleSheet)
```

### Constraint decorators
e.g. The `height(_:)` and `width(_:)` DSL operators can now be implemented as:

```swift
private let heightConstraintKey = ViewStorage.Key<NSLayoutConstraint>()

struct HeightDecorator<R: Renderable>: Renderable {
    let base: R
    let height: CGFloat

    // init decl...

    func render(in view: R.View) {
        base.render(in: view)
    }

    func didMount(to view: View, storage: ViewStorage) {
        base.didMount(to: view, storage: storage)
        storage[heightConstraintKey] = view.heightAnchor
            .constraint(equalToConstant: height)
            .activated()
    }

    func willUnmount(from view: View, storage: ViewStorage) {
        base.willUnmount(from: view, storage: storage)
        storage[heightConstraintKey]?.isActive = false
    }
}
```

PR #164, opened 2019-05-05
Collapse `ComponentLifecycleAware` into `Renderable`.

andersio:
Collapse the two existing component lifecycle method requirements back into `Renderable`. Default implementations are supplied. The fleet will be joined by the two new hooks `willMount` and `didMount` in #162.

The two methods now also receive the view, making it is functionally equivalent to `ViewLifecycleAware` but requiring no protocol conformance from the view type.

The only caveat is that since composite components in our monorepo at the moment do not manage the lifecycle of their children as per the updated contract in #165 (which they currently don't), `ViewLifecycleAware` is still kept for compatibility but marked as deprecated.

```swift
protocol Renderable {
    // ... other req

    func willDisplay(_ view: View)
    func didEndDisplaying(_ view: View)
}
```

This should improve discoverability. By promoting them to be explicit part of the single `Renderable` contract it hopefully inspires implementation of composites not to forget them calling them, since not doing so now breaks the perceived `Renderable` lifecycle.


sergdort (approved):
👍

mluisbrown (approved):

PR #166, opened 2019-05-07
Adds AnyRenderableConvirtable

TheAdamBorek:
I'd like to add `AnyRenderableConvirtable` into Bento.

## Why?
In our main project, we need to have an array of `AnyRenderable` to be able to call `.stack(...)` method. Because of that, every element of the array has to have `.asAnyRenderable()` called at the end of its declaration. I'm just too lazy to do so every time.

With `AnyRenderableConvirtable` the `Stack` component can be responsible to call `asAnyRenderable()` on it's children :)  

```
extension Array where Element == Optional<AnyRenderableConvirtable> { ... }
```

### DIFF
--- a/Bento/Renderable/AnyRenderable.swift
+++ b/Bento/Renderable/AnyRenderable.swift
@@ -1,5 +1,15 @@
 import UIKit
 
+public protocol AnyRenderableConvertible {
+    func asAnyRenderable() -> AnyRenderable
+}
+
+public extension AnyRenderableConvertible where Self: Renderable {
+    func asAnyRenderable() -> AnyRenderable {
+        return AnyRenderable(self)
+    }
+}
+
 public struct AnyRenderable: Renderable {
     /// The runtime view type of the wrapped `Renderable`.
     public var viewType: NativeView.Type {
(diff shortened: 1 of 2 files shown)


mluisbrown (approved):

NSMyself (approved):
Nice work mate

Bento/Renderable/AnyRenderable.swift
```diff
@@ -1,5 +1,15 @@
 import UIKit
 
+public protocol AnyRenderableConvertible {
+    func asAnyRenderable() -> AnyRenderable
+}
+
+public extension AnyRenderableConvertible where Self: Renderable {
```
inamiy, line 7: This is essentially equal to `extension Renderable` .

inamiy (commented):
I'm not quite sure if having intermediate public protocol will ever be useful in any cases... 🙄 

```swift
public protocol Xxxable: AnyXxxConvertible
```

Essentially, we can fill in any `Xxx` here when making protocol (and guide to its type-erasure).
But we don't normally create this intermediate protocol.

Issue #163, opened 2019-05-05
Dependency cycle


eimantas:
Trying to install Bento 0.4 using carthage:

```console
% carthage update bento --platform ios 
*** Cloning Bento
*** Fetching ReactiveCocoa
*** Fetching TinyConstraints
*** Cloning FlexibleDiff
*** Fetching ReactiveSwift
*** Cloning Result
*** Cloning xcconfigs
*** Cloning Box
*** xcodebuild output can be found in /var/folders/w3/d5j4zf_13xnf7ppx7qlps8wh0000gn/T/carthage-xcodebuild.GupBlD.log
The dependency graph contained a cycle:
TinyConstraints: 
ReactiveCocoa: ReactiveSwift
% swift --version
Apple Swift version 5.0.1 (swiftlang-1001.0.82.4 clang-1001.0.46.5)
Target: x86_64-apple-darwin18.5.0
% carthage version
0.33.0
% 
```

Cartfile looks like this:

```console
github "ReactiveCocoa/ReactiveCocoa" ~> 10.0.0
github "roberthein/TinyConstraints" ~> 4.0.0
github "babylonpartners/Bento" ~> 0.4
```

TheAdamBorek:
Hey @eimantas. We're on it. Meanwhile, can you check if this branch `adam/fix-carthage-build` is building fine on your machine? :)

eimantas:
Nope, same error:

```console
% carthage update Bento --platform ios --cache-builds
*** Fetching TinyConstraints
*** Fetching Bento
*** Fetching ReactiveCocoa
*** Fetching ReactiveSwift
*** Fetching xcconfigs
*** Checking out FlexibleDiff at "0.0.8"
*** Checking out Bento at "bae0eaf88fe53d0062e7bc518eb5b6719204d50d"
*** xcodebuild output can be found in /var/folders/w3/d5j4zf_13xnf7ppx7qlps8wh0000gn/T/carthage-xcodebuild.0fB3Hy.log
The dependency graph contained a cycle:
TinyConstraints: 
ReactiveCocoa: ReactiveSwift
FlexibleDiff: xcconfigs
Bento: FlexibleDiff
xcconfigs: 
% 
```

Cartfile:

```console
github "ReactiveCocoa/ReactiveCocoa" ~> 10.0.0
github "roberthein/TinyConstraints" ~> 4.0.0
github "babylonpartners/Bento" "adam/fix-carthage-build"
```

[...]

PR #168, opened 2019-05-07
Fixes carthage build

TheAdamBorek:
I've made the `BentoKit` scheme not-shared so we don't longer need `ReactiveCocoa` in `Cartfile` (as @mluisbrown suggested).
It should solve https://github.com/Babylonpartners/Bento/issues/163 

It's a hotfix only. We should remove BentoKit completely from the repo in a future release.

The `hotfix/0.4.1` branch has been created form `master`.

EDIT:
I also had to update CircleCi config


mluisbrown:
Is the plan to move `BentoKit` to a separate repo?

TheAdamBorek:
I think "yes". However, the fact you asked the question I'm not sure now 😅

mluisbrown:
Does CI only work with shared schemes? It’s a pity we have to lose CI for BentoKit with this fix for Carthage :/

TheAdamBorek:
yeah, you are right @mluisbrown. I was saying to myself "I can delete it because we delete the BentoKit".... but we still use it internally :/ 

In that case I don't know if we should do this. I don't know CircleCI so much, however it has reported that test has failed for BentoKit in this PR.

```
#!/bin/bash -eo pipefail
script/test iphonesimulator "platform=iOS Simulator,name=iPhone 6,OS=12.2" BentoKit
^D^Dxcodebuild: error: The workspace named "Bento" does not contain a scheme named "BentoKit". The "-list" option can be used to find the names of the schemes in the workspace.
Exited with code 65
```

Maybe we should just set `ReactiveCocoa` to version `10.0.0` instead of `9.0.0` until we move BentoKit into a new repository 🤔. However, `ReactiveCocoa 10.0.0` removed `Result` dependency in favour to use Swift's Result. I have a bad feeling it's not backward compatible change :/

mluisbrown:
Having thought about this some more, my view is that we should not make this change at all. The issue was conflicting versions of RAC. The user had `10.0.0` in their `Cartfile`, whereas `BentoKit` was requresing `9.0.0`. 

I think the correct "resolution" to the issue would have been to inform the user that, for the time being at least, if you use Bento, you have to use RAC 9.0.0.

If we keep BentoKit in this repo, then this will be, IMO, a minor issue for some users. You will just have to always use the same version of RAC that BentoKit requires 🤷‍♂ 

If we move BentoKit to a separate repo, that would solve the problem entirely.

mluisbrown (changes requested):
Requesting changes so this doesn't get accidentally merged. See my comment.

PR #169, opened 2019-05-07
Change the access level of variables in BoxViewController

gtsiap:
These variables can be useful if we need to subclass `BoxViewController`.

### DIFF
--- a/BentoKit/BentoKit/Screen/BoxViewController.swift
+++ b/BentoKit/BentoKit/Screen/BoxViewController.swift
@@ -7,8 +7,8 @@ import UIKit
 open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, Appearance>
 : UIViewController where ViewModel.State == Renderer.State, ViewModel.Action == Renderer.Action, Renderer.Appearance == Appearance {
     public let viewModel: ViewModel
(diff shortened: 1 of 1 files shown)


TheAdamBorek (approved):
It's still `let` so I'm fine with that. I think @andersio can also look at this

BentoKit/BentoKit/Screen/BoxViewController.swift
```diff
@@ -7,8 +7,8 @@ import UIKit
 open class BoxViewController<ViewModel: BoxViewModel, Renderer: BoxRenderer, Appearance>
 : UIViewController where ViewModel.State == Renderer.State, ViewModel.Action == Renderer.Action, Renderer.Appearance == Appearance {
     public let viewModel: ViewModel
-    private let rendererConfig: Renderer.Config
-    private let appearance: Property<Appearance>
+    public let rendererConfig: Renderer.Config
+    public let appearance: Property<Appearance>
```
sergdort, line 11: Why do we need it to be `public`?

diegopetrucci (changes requested):
Do we already have a need for this, or is it speculative?

diegopetrucci (approved):
Okay, just saw the referenced PR.

PR #171, opened 2019-05-11
Fix rendering issue with `PreSizingLayoutPassRequiring`

### LINKED ISSUE #170: Rendering bug with components conforming to `PreSizingLayoutPassRequiring`
A `Component` conforming to `PreSizingLayoutPassRequiring` exhibits strange rendering behaviour, appearing to render at zero height and then re-size to the correct height on every render pass. This can be seen in the SignUp Example:

BentoSizeCachingIssue

The Gender field is using `DetailedDescripton` which conforms to `PreSizingLayoutPassRequiring`. If I remove this conformance, the rendering glitch goes away.

I suspect that this is somehow related to the recent changes relating to size caching, as this only started to occur recently (on the `develop` branch, btw).

Seems related to #80 

cc: @andersio

mluisbrown:
Only set the `bounds.size.width` from the target size, which is all that is needed to calculate the cell height. Fixes #170

Setting the `bounds.size` (including the `height`) resets the height of the cell leading to the cell being always rendered first with zero height, and then the correct height.

### DIFF
--- a/Bento/Views/RequiringPreSizingLayoutPass.swift
+++ b/Bento/Views/RequiringPreSizingLayoutPass.swift
@@ -37,6 +37,12 @@ extension UIView {
 
     func triggerPresizingLayoutPassIfNeeded(forTargetSize size: CGSize) {
         if needsPresizingLayoutPass {
+            // NOTE: the target size passed in always has a height of zero
+            // when called from `systemLayoutSizeFitting(_ targetSize: CGSize)`
+            // All we care about is the width. Setting the height to zero
+            // would make the cell render with zero height first then animate
+            // up to the correct height on every render pass.
+            // `layoutIfNeeded()` will take of setting the correct height.
             bounds.size.width = size.width
             layoutIfNeeded()
         }
--- a/Bento/Views/RequiringPreSizingLayoutPass.swift
+++ b/Bento/Views/RequiringPreSizingLayoutPass.swift
@@ -37,7 +37,7 @@ extension UIView {
 
     func triggerPresizingLayoutPassIfNeeded(forTargetSize size: CGSize) {
         if needsPresizingLayoutPass {
-            bounds.size = size
+            bounds.size.width = size.width
             layoutIfNeeded()
         }
     }


sergdort (approved):

Bento/Views/RequiringPreSizingLayoutPass.swift
```diff
@@ -37,7 +37,13 @@ extension UIView {
 
     func triggerPresizingLayoutPassIfNeeded(forTargetSize size: CGSize) {
         if needsPresizingLayoutPass {
-            bounds.size = size
+            // NOTE: the target size passed in always has a height of zero
+            // when called from `systemLayoutSizeFitting(_ targetSize: CGSize)`
+            // All we care about is the width. Setting the height to zero
```
TheAdamBorek, line 42: > All we care about is the width.

Is that true it case of UICollectionView? 🤔

mluisbrown: Well, `systemLayoutSizeFitting(_ targetSize: CGSize)` is only overridden for `TableViewContainerCell` and `TableViewHeaderFooterView`, it's not needed for `UICollectionView`.

However `AnyRenderable.rendered(size: CGSize, inheritedMargins: UIEdgeInsets)` also calls `triggerPresizingLayoutPassIfNeeded`, and this applies to `UICollectionView` too.

However, the whole premise of `PreSizingLayoutPassRequiring` is *height* calculation which is dependent on a known width. From the comments for this protocol:

```swift
/// Declare that the view requires an extra layout pass before Auto Layout item sizing happens, so that height
/// calculation for components containing free-form UI elements, e.g. text labels, would be correct.
```

Also from the comments:

```swift
/// - important: The mechanism is always enabled for Bento-enabled `UITableView`, both vanilla and size-caching. For
///              `UICollectionView` however, it is available only for size-caching `UICollectionView`, since
///              `UICollectionView` can be flexibly laid out and thus Bento does not have sufficient knowledge when
///              working with vanilla `UICollectionView`.
```

Basically, for `UICollectionView` you have to depend on auto-layout, because your cells might have a fixed width *and* height.

TheAdamBorek (approved):

PR #174, opened 2019-05-14
Fix TD's detail label becoming crushed due to a large title

NSMyself:
This PR fixes a bug in which large titles inside a TitledDescription would shrink the detail label up to a point that it would not longer be readable. 

In order to fix this, the content compression resistance priority of the label itself was simply lowered from `defaultHigh` to `defaultLow`.


mluisbrown:
Can you please make the same fix to `DetailedDescription` in `Bento`?
Also, IMO we should take this opportunity to remove `TitledDescription` from `BentoKit` and move it to `BabylonUI`.

mluisbrown (approved):

PR #175, opened 2019-05-23
Refactor StackView's stylesheet to avoid invalid configuration

dmcrodrigues:
## Why?

There's a series of message logs being printed like this one:

```
<CATransformLayer: 0x7f9cd2a890e0> - changing property shadowOffset in transform-only layer, will have no effect
```

And this happens due the different nature of `UIStackView`s.

> The UIStackView is a nonrendering subclass of UIView; that is, it does not provide any user interface of its own. Instead, it just manages the position and size of its arranged views. As a result, some properties (like backgroundColor) have no effect on the stack view. Similarly, you cannot override layerClass, draw(_:), or draw(_:in:).

StackView's stylesheet cannot inherit from `ViewStyleSheet` because the majority of the properties do not apply to it, it should instead focus in managing the properties characteristic of `UIStackView` which goes from the `axis` to the `distribution` but also includes some more general properties like `layoutMargins` and `clipToBounds`. 

It's specially problematic because stylesheets contains all the default values to be applied and ensure the views are reset correctly, e.g. before being reused, which produces the problem even when the consumer doesn't try to set any of this invalid properties.

## How?

`StackViewStyleSheet` is now a base stylesheet focused in configuring the properties applicable to any stack view. Because we also have a `BaseStackView` which extends `UIStackView` with more functionality like having a background color we have another stylesheet `BaseStackViewStyleSheet` which inherits from `StackViewStyleSheet` and extends it with all the extra properties valid for any `BaseStackView`.


mluisbrown (approved):
Finally 🎉

diegopetrucci (approved):

Bento/StyleSheets/StackViewStyleSheet.swift
```diff
@@ -1,31 +1,43 @@
 import UIKit
 
-open class StackViewStyleSheet: ViewStyleSheet<UIStackView> {
+open class StackViewStyleSheet<View: UIStackView>: StyleSheetProtocol {
```
inamiy, line 9: Can we have a doc-comment for this class as mentioned in PR body?
(especially why we don't inherit `ViewStyleSheet` )

dmcrodrigues: @inamiy I've included documentation to explain it.

dmcrodrigues:
This will be a breaking change in the public API, do we have any strategy for this kind of changes already in place? Write a migration guide and release a new major version?

AliSoftware:
> This will be a breaking change in the public API, do we have any strategy for this kind of changes already in place? Write a migration guide and release a new major version?

Migration Guide would be nice, but as far as versioning is concerned, according to semver.org (which is what CP follows too), the rules don't apply on pre-1.0 versions: A library version `0.x.y` (like Bento) is supposed to mean that the API is not stable yet and that the API can be modified/adjusted at any time up until 1.0.

dmcrodrigues:
Yes @AliSoftware, my main concern is how this would affect our consumers mostly, not sure if we have or want to have something in place for this kind of moments or if we want to keep it simple and agile until we reach `1.0`.

Bento/StyleSheets/StackViewStyleSheet.swift
```diff
@@ -1,31 +1,49 @@
 import UIKit
 
-open class StackViewStyleSheet: ViewStyleSheet<UIStackView> {
+/// A stylesheet which can configure any instance of `UIStackView`.
+///
+/// `UIStackView` is a nonrendering subclass of `UIView` and because of that we cannot inherit from `ViewStyleSheet` due
+/// the majority of the properties of a view cannot be applied to a stack view. This stylsheet focus in managing all the
+/// configuration characteristic of any `UIStackView` which goes from the `axis` to `distribution` but also includes
+/// some more general properties like `layoutMargins` and `clipToBounds`.
```
inamiy:
```suggestion
/// some more general properties like `layoutMargins` and `clipsToBounds`.
```

inamiy (approved):
@dmcrodrigues 
Sorry for late review, but approved 👍 
One nit suggestion added, so pls merge after fixed.

PR #176, opened 2019-05-28
Remove TitledDescription, tests and snapshots, and fix some compiler warnings

mluisbrown:
As has previously been discussed, it makes no sense for the `TitledDescription` component, which depends on ReactiveSwift, to continue as the only component in `BentoKit`, now that all the other components have been moved to Bento.

This PR removes `TitledDescription` from BentoKit.


diegopetrucci (approved):

NSMyself (approved):
👍

Bento/Components/Button.swift
```diff
@@ -182,7 +182,7 @@ extension Component.Button {
 }
 
 public extension Component.Button {
-    public final class StyleSheet: InteractiveViewStyleSheet<View> {
+    final class StyleSheet: InteractiveViewStyleSheet<View> {
         public let button: ButtonStyleSheet
```
inamiy, line 186: These `public`s will be confusing, so it's better to also remove all of them.

mluisbrown: I just removed the ones that were causing compiler warnings.

inamiy: Never mind, it's `public extension` which is already confusing... 🙄

PR #177, opened 2019-05-29
Make `Section.init(id:items:supplements:)` public

inamiy:
This PR exposes the most basic initializer to the public so that optional `header` / `footer` can be easily set and filtered.

For example, filtering using `compactMapValues`:

```swift
let section = Section(id: ..., items: ..., supplements: [
    .header: maybeHeader()
    .footer: maybeFooter()
].compactMapValues { $0 })
```

Note that current `init`s are overloaded with taking non-optional `header` / `footer` arguments, which makes hard to set optional value as a filterable type.

Making this initializer public makes sense as long as `public func Section.adding` is also public.

### Alternative consideration

```diff
-    public init<Header: Renderable, Footer: Renderable>(id: SectionID, header: Header, footer: Footer, items: [Item] = []) {
+    public init<Header: Renderable, Footer: Renderable>(id: SectionID, header: Header? = nil, footer: Footer? = nil, items: [Item] = []) {
```

Though above code might work, there may still exist ambiguity with other overloaded `init`s, so I decided not to modify as so.

### DIFF
--- a/Bento/Bento/Section.swift
+++ b/Bento/Bento/Section.swift
@@ -35,7 +35,7 @@ public struct Section<SectionID: Hashable, ItemID: Hashable> {
                             .footer: AnyRenderable(footer)]
     }
 
-    internal init(id: SectionID, items: [Item], supplements: [Supplement: AnyRenderable]) {
+    public init(id: SectionID, items: [Item], supplements: [Supplement: AnyRenderable]) {
         self.id = id
         self.items = items
         self.supplements = supplements


Bento/Bento/Section.swift
```diff
@@ -35,7 +35,7 @@ public struct Section<SectionID: Hashable, ItemID: Hashable> {
                             .footer: AnyRenderable(footer)]
     }
 
-    internal init(id: SectionID, items: [Item], supplements: [Supplement: AnyRenderable]) {
+    public init(id: SectionID, items: [Item], supplements: [Supplement: AnyRenderable]) {
```
sergdort, line 38: This will allow to put `supplements` for `UITableView` which can have only header and footer

inamiy: Right, but we already have `public func adding<R: Renderable>(_ supplement: Supplement, _ component: R) -> Section`

sergdort: True

inamiy:
BTW current solution is to use `var` and `adding` which obviously isn't great:

```swift
var section = Section(id: ...)
if let header = maybeHeader {
    section = section.adding(.header, ...)
}
if let footer = maybeFooter {
    section = section.adding(.footer, ...)
}

return section
    |---+ Node(...)
    |---+ Node(...)
```

mluisbrown:
You can use `adding` without a `var`:

```swift
let section = Section(id: ...)
let header: HeaderComponent? = maybeHeader

return header.map { section.adding(.header, $0) } ?? section
```

But I agree it's a bit messy.

mluisbrown (approved):

andersio (approved):
It was kept internal only because back in the time `AnyRenderable` was Bento internal.

PR #179, opened 2019-05-30
[CNSMR-1713] Custom Input should work with AnyRenderable.

andersio:
https://babylonpartners.atlassian.net/browse/CNSMR-1713

Bento Custom Input operator should work when decorating the type-erased AnyRenderable. This is currently crashing since statically speaking `AnyRenderable.View` is not the same as its wrapping value's `View` type.

### DIFF
--- a/Bento/Views/BentoReusableView.swift
+++ b/Bento/Views/BentoReusableView.swift
@@ -1,20 +1,23 @@
-protocol BentoReusableView: NativeView {
-    var containedView: NativeView? { get set }
-    var contentView: NativeView { get }
-    var component: AnyRenderable? { get set }
+public protocol ViewStorageOwner: NativeView {
     var storage: [StorageKey: Any] { get set }
 }
 
-struct StorageKey: Hashable {
+protocol BentoReusableView: ViewStorageOwner {
+    var containedView: NativeView? { get set }
+    var contentView: NativeView { get }
+    var component: AnyRenderable? { get set }
+}
+
+public struct StorageKey: Hashable {
     let component: Any.Type
     let identifier: ObjectIdentifier
 
-    func hash(into hasher: inout Hasher) {
+    public func hash(into hasher: inout Hasher) {
         hasher.combine(ObjectIdentifier(component))
         hasher.combine(identifier)
     }
 
-    static func == (lhs: StorageKey, rhs: StorageKey) -> Bool {
+    public static func == (lhs: StorageKey, rhs: StorageKey) -> Bool {
         return lhs.component == rhs.component && lhs.identifier == rhs.identifier
     }
 }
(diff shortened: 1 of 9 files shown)


Bento/Views/BentoReusableView.swift
```diff
@@ -1,20 +1,23 @@
...
+
+protocol BentoReusableView: ViewStorageOwner {
     var containedView: NativeView? { get set }
     var contentView: NativeView { get }
     var component: AnyRenderable? { get set }
-    var storage: [StorageKey: Any] { get set }
 }
 
-struct StorageKey: Hashable {
+public struct StorageKey: Hashable {
```
inamiy, line 9: Do we need public here? (same for `ViewStorageOwner` )

NSMyself (approved):
🎉

mluisbrown (approved):

Bento/Views/BentoReusableView.swift
andersio: Yes, because there are composites in BabylonUI and they should call these lifecycle methods too (although they don’t) & therefore these are needed for them to be able to call `didMount` and `willUnmount`.

inamiy: Then, do we need `ViewStorage.init` as `public` ?

inamiy:
BTW are similar fixes also needed for other compositions e.g. `DeletableComponent` ?
(If so, though, fixes can go to next PR)

inamiy:
I'm starting to feel that adding `ViewStorageOwner` for properly handling `didMount` / `willUnmount` is a separate issue from this PR's objective.
It is bringing another complexity since #162 , and as I also questioned in https://github.com/Babylonpartners/Bento/pull/179#issuecomment-497364809 , there may need more relevant fixes.

There's also a question about if we can simplify code as follows, eventually removing `ViewStorage` in future:
https://gist.github.com/inamiy/4d3d8a9434cb7bcab4baf370fac647d0
(but I don't want to discuss this further in this PR)

Thus, can you split the PR to only focus on `CustomInput` fix for now?
Or, I can't approve until all irrelevant questions are discussed here and answered.

Bento/Views/BentoReusableView.swift
andersio: Noice catch.

andersio:
~~@inamiy Adding `ViewStorageOwner` is essential. Without it, either composite components conform their views to `BentoReusableView` (larger leap), or none of them can construct `ViewStorage` (`BentoReusableView` is Bento internal).~~

andersio:
Removed `ViewStorageOwner`. Extend instead `BentoReusableView` to support the use cases. I suppose making `BentoReusableView` public might be a better choice after all, but it is not done in this PR.

NSMyself (approved):

NSMyself (approved):

Bento/Views/BentoReusableView.swift
inamiy: With 0b32de8 , this type doesn't need to be `public` for now until making `BentoReusableView` public.

But I guess it will become `public` again soon anyway, so I will ignore this.

inamiy (approved):
0b32de8 looks simpler now 👍

PR #182, opened 2019-06-04
Add async-`CustomInput` support

inamiy:
This PR is an experiment of **asynchronously displaying `CustomInput`** after cell tap.
Example demo can be found in `SignUp` screen.

This feature is needed when `OptionPicker` 's data is not ready yet and requires async fetch before `CustomInput` presentation.
It is triggered by state change and `focusesOnFirstDisplay` is set. 

The implementation is complete, so please review and tell me this is the right way to go.
**(Please don't review about example code. It will be deleted afterwards)**

## Changes

- ad713ac Added `func customInputImmediately(when:input:)`
- Added `focusesOnFirstDisplay`
- Allow `CustomInputComponent.ComponentView` to present `FocusToolbar` without back/forward buttons when it is not focus-eligible (due to asynchrony).

## TODO

- [x] Organize example

### DIFF
--- a/Bento/Decorators/CustomInputComponent.swift
+++ b/Bento/Decorators/CustomInputComponent.swift
@@ -3,14 +3,16 @@ import UIKit
 extension Renderable {
     public func customInput(
         _ input: CustomInput,
-        contentStatus: FocusEligibility.ContentStatus = .empty,
-        highlightColor: UIColor? = UIColor(red: 239/255.0, green: 239/255.0, blue: 244/255.0, alpha: 1)
+        focusEligibility: FocusEligibility = .eligible(.empty),
+        highlightColor: UIColor? = UIColor(red: 239/255.0, green: 239/255.0, blue: 244/255.0, alpha: 1),
+        focusesOnFirstDisplay: Bool = false
     ) -> AnyRenderable {
         return CustomInputComponent(
             source: self,
             customInput: input,
-            contentStatus: contentStatus,
-            highlightColor: highlightColor
+            focusEligibility: focusEligibility,
+            highlightColor: highlightColor,
+            focusesOnFirstDisplay: focusesOnFirstDisplay
         ).asAnyRenderable()
     }
 }
@@ -19,27 +21,36 @@ struct CustomInputComponent: Renderable, Focusable {
     let customInput: CustomInput
     let focusEligibility: FocusEligibility
     let highlightColor: UIColor?
+    let focusesOnFirstDisplay: Bool
+
     let base: AnyRenderable
 
     init<Base: Renderable>(
         source: Base,
         customInput: CustomInput,
-        contentStatus: FocusEligibility.ContentStatus,
-        highlightColor: UIColor?
+        focusEligibility: FocusEligibility,
+        highlightColor: UIColor?,
+        focusesOnFirstDisplay: Bool
     ) {
         self.customInput = customInput
         self.highlightColor = highlightColor
         self.base = AnyRenderable(source)
-        self.focusEligibility = .eligible(contentStatus)
+        self.focusEligibility = focusEligibility
+        self.focusesOnFirstDisplay = focusesOnFirstDisplay
     }
 
     func render(in view: ComponentView) {
         view.inputNodes = customInput
+        view.isFocusEnabled = focusEligibility.isEligible(skipsPopulatedComponents: false)
         view.highlightingGesture.didTap = .manual
         view.highlightingGesture.highlightColor = highlightColor
         view.highlightingGesture.stylingView = view.containedView
 
         view.bind(base)
+
+        if focusesOnFirstDisplay && view.canBecomeFirstResponder && !view.isFirstResponder {
+            _ = view.becomeFirstResponder()
+        }
     }
 
     func willDisplay(_ view: CustomInputComponent.ComponentView) {
@@ -64,6 +75,8 @@ extension CustomInputComponent {
                 }
             }
         }
+
+        fileprivate(set) var isFocusEnabled: Bool = true
 
         var customInputView: InputView?
         var focusToolbar: FocusToolbar?
@@ -103,7 +116,7 @@ extension CustomInputComponent {
         public override func becomeFirstResponder() -> Bool {
             if let nodes = inputNodes {
                 customInputView = InputView()
-                focusToolbar = FocusToolbar(view: self)
+                focusToolbar = FocusToolbar(view: self, isFocusEnabled: isFocusEnabled)
                 customInputView!.update(nodes)
             }
 
--- a/Bento/Decorators/CustomInputComponent.swift
+++ b/Bento/Decorators/CustomInputComponent.swift
@@ -1,24 +1,43 @@
 import UIKit
 
 extension Renderable {
+    /// Sets up `customInput` to prepare for presention when user taps `self`.
     public func customInput(
         _ input: CustomInput,
-        focusEligibility: FocusEligibility = .eligible(.empty),
-        highlightColor: UIColor? = UIColor(red: 239/255.0, green: 239/255.0, blue: 244/255.0, alpha: 1),
-        focusesOnFirstDisplay: Bool = false
+        contentStatus: FocusEligibility.ContentStatus = .empty,
+        highlightColor: UIColor? = UIColor(red: 239/255.0, green: 239/255.0, blue: 244/255.0, alpha: 1)
     ) -> AnyRenderable {
         return CustomInputComponent(
             source: self,
             customInput: input,
-            focusEligibility: focusEligibility,
+            focusEligibility: .eligible(contentStatus),
             highlightColor: highlightColor,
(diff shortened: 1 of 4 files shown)


inamiy:
ad713ac 
Simplified code to `func customInputImmediately(when:input:)`.
Previous tweaks in `func customInput` is reverted.

Example/SignUp/SignUpPresetner.swift
```diff
@@ -50,6 +50,16 @@ final class SignUpPresenter {
         navigator?.showAlert(title: "Did sign up", message: "\(state)")
     }
 
+    /// - note: Simulates asynchronous picker presentation
```
RuiAAPeres: Why do we need this comment? 1) It's the example app. 2) At line 57 it's obvious that it's async. 🤔

Example/SignUp/SignUpPresetner.swift
```diff
@@ -141,3 +152,15 @@ final class SignUpPresenter {
         }
     }
 }
+
+/// Async gender picker state.
```
RuiAAPeres: Not sure how this comment is relevant as well. The enum by itself has no relationship with where it is used.

Example/SignUp/SignUpPresetner.swift
inamiy: Example code will be deleted after having some reviews.
I need to clarify my impl is the right approach first.

Bento/Decorators/CustomInputComponent.swift
```diff
@@ -9,37 +10,66 @@ extension Renderable {
...
+    /// and also presents it immediately when `state` is `.some`.
+    ///
+    /// - important: This method is useful when `customInput` needs to be displayed asynchronously after state change.
+    /// - note: Due to asynchronous presentation, `focusEligibility` is not supported.
+    public func customInputImmediately<State>(
+        when state: State?,
+        input: (State) -> CustomInput,
+        highlightColor: UIColor? = UIColor(red: 239/255.0, green: 239/255.0, blue: 244/255.0, alpha: 1)
+    ) -> AnyRenderable {
+        return CustomInputComponent(
```
dmcrodrigues, line 28: Do we really need to change `CustomInputComponent` to support this? Why can't we just return `self` when state is `nil`? At least I think this should be done the same way that we build boxes where sections and/or nodes are conditionally composed depending on a certain condition and here I think we could follow the same principle, if the `state` is `nil` it means there's no `CustomInput` so we don't need a `CustomInputComponent` but once that changes we can then build a `CustomInputComponent` to display that input.

I'm not sure if I'm missing some detail.

Bento/Decorators/CustomInputComponent.swift
```diff
@@ -9,37 +10,66 @@ extension Renderable {
...
+            focusesOnFirstDisplay: false
+        ).asAnyRenderable()
+    }
+
+    /// Sets up `customInput` to prepare for presention when user taps `self`,
+    /// and also presents it immediately when `state` is `.some`.
+    ///
+    /// - important: This method is useful when `customInput` needs to be displayed asynchronously after state change.
+    /// - note: Due to asynchronous presentation, `focusEligibility` is not supported.
+    public func customInputImmediately<State>(
```
dmcrodrigues: Also, I find the name confusing because this won't show immediately since it first triggers some action and only then displays the custom input. I think using the same name would be better because the parameters are clear enough to distinguish between both behaviours, either you have your input available or you are conditionally waiting for some state to get it.

Bento/Decorators/CustomInputComponent.swift
inamiy: `state = nil` doesn't mean we don't use `customInput` at all.
It's rather the same behavior as we already have: "Show customInput when user tapped"

On the contrary, when `state` exists, `customInput` shows **immediately** when re-rendered.
It also holds the same property  "show customInput when user tapped" as well.

inamiy: Maybe I should rename method e.g. `func customInput(immediatelyWhen:...)` so that "immediately" belongs to state existence.

Example/SignUp/SignUpPresetner.swift
inamiy: Same here https://github.com/Babylonpartners/Bento/pull/182#discussion_r292845574

Bento/Decorators/CustomInputComponent.swift
inamiy: https://github.com/Babylonpartners/Bento/pull/182#discussion_r292862688

> `func customInput(immediatelyWhen:...)`

How about this?

Bento/Decorators/CustomInputComponent.swift
dmcrodrigues: > state = nil doesn't mean we don't use customInput at all.
It's rather the same behavior as we already have: "Show customInput when user tapped"

I'm a bit confused, in your example you tap `Gender` and there's no input being presented, you display an activity indicator as accessory and after 1 sec. your state changes and then it shows the option picker, why do we need to have a custom input before the change of state?

inamiy: > why do we need to have a custom input before the change of state

That's good question!
IIRC I tried the same approach by removing `customInput` only when loading.
But this caused Bento diffing animation for the entire "gender cell" that I wanted to avoid.
So, keeping `customInput` throughout the handling is (probably) required to avoid such animations.

dmcrodrigues: Oh yes it does that because we are swapping from one component to another one if we do that 😕

I don't see any way to avoid that animation but maybe @andersio has an idea.

inamiy: There could be some diffing optimization, but I think it's still OK to always keep `customInput` during loading.
It might still be able to tap though (which will be a bug), I can add `isEnabled` property to prevent that, so that we don't need to worry about removal animation of `customInput`.

inamiy: (For sure diffing optimization will be more nice, but we won't probably invest too much here)

Bento/Decorators/CustomInputComponent.swift
inamiy: bd6df2f Renamed to `func customInput(immediatelyWhen:...)` so that it becomes more clear that `customInput` is always placed as a virtual node.

inamiy:
In PR party, @andersio came up with an idea of showing loading indicator inside keyboard so that focus eligibility (keyboard switching) can still work and has a better user experience.

While I agree on this UX improvement, this PR is still orthogonal that we want to have some functionality to present the keyboard immediately, e.g. SwiftUI's `presentation(_: Modal?)`  driven by state change to set `Modal`.

So, we concluded that this PR will be reviewed separately without considering about loading-keyboard.

Bento/Decorators/CustomInputComponent.swift
andersio: #146 is meant to tackle that. But I'd be surprised that there is any deleting animation. It should be a subtle cross-fade.

Bento/Decorators/CustomInputComponent.swift
```diff
@@ -9,37 +10,66 @@ extension Renderable {
...
+        ).asAnyRenderable()
+    }
+
+    /// Sets up `customInput` to prepare for presention when user taps `self`,
+    /// and also presents it immediately when `state` is `.some`.
+    ///
+    /// - important: This method is useful when `customInput` needs to be displayed asynchronously after state change.
+    /// - note: Due to asynchronous presentation, `focusEligibility` is not supported.
+    public func customInput<State>(
+        immediatelyWhen state: State?,
```
andersio: It could've been written as simply accepting `input: CustomInput?` for the illustrated use cases. Not sure what value the indirection adds here.

andersio (changes requested):
Ditto

Bento/Decorators/CustomInputComponent.swift
inamiy: There is already an existing `func customInput(_ input: CustomInput, ...)` , so removing this indirection from new method makes harder to distinguish the differences between 1. `customInput(input)`, 2. `customInput(.some(input))`, 3. `customInput(.none)`.

But I do agree that new method with `CustomInput?` argument is simpler and works like SwiftUI, so changing existing method is probably more preferred here, but I believe it's out of this PR's scope.

If you don't mind overlapping the similar method signature, I will change it.
What do you think?

andersio: OK... `autodisplayingCustomInput(_:)`? I think the behaviour is different enough that giving it a different base name is probably better.

inamiy: Renamed in ef12cd7 .

inamiy:
f789bcc Reverted example code.

andersio (approved):

dmcrodrigues (approved):

PR #184, opened 2019-06-05
Add `extension String: Option`

inamiy:
Added `extension String: Option` since there is not much reason not to.


AliSoftware (approved):
LGTM

dmcrodrigues (approved):

PR #185, opened 2019-06-11
[GW-667] Remove call to `didRenderBox` in `UICollectionView.render`.

mluisbrown:
`didRenderBox` sends a `#selector(FocusableView.neighboringFocusEligibilityDidChange)` up the responder chain, which eventually would cause a `UIKeyboardWillChangeFrameNotification` to be sent to the `UICollectionView` which is observing these notifications via `NotificationCenter.reactive.notifications(forName name: , object: )`.

If this happens as a consequence of a text field calling `resignFirstResponder`, which **also** causes a `UIKeyboardWillChangeFrameNotification` to be sent, then you can get a recursive lock exception as `NotificationCenter.reactive.notifications(forName name: , object: )` is trying to `send` recursively to the same `Observer`.

Here is an example stack trace of this happening:
image

The workaround is to remove the call to `didRenderBox` in `UICollectionView.render`. This would have the side effect of breaking focus support in `UICollectionView` screens, but there appears to be no use of this feature currently.

### DIFF
--- a/Bento/Bento/UICollectionViewExtensions.swift
+++ b/Bento/Bento/UICollectionViewExtensions.swift
@@ -28,7 +28,6 @@ extension UICollectionView {
     public func render<SectionID, ItemID>(_ box: Box<SectionID, ItemID>, animated: Bool = true, completion: (() -> Void)? = nil) {
         let adapter: CollectionViewAdapterBase<SectionID, ItemID> = getAdapter()
         adapter.update(sections: box.sections, animated: animated, completion: completion)
-        didRenderBox()
     }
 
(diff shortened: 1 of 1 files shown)


MichalTKwiecien (approved):
Hope it won't break my other PR 🙈 https://github.com/Babylonpartners/babylon-ios/pull/7782

mluisbrown:
Marked this as Don't Merge for now as I try a different approach.

mluisbrown:
Ok, this is not required. There is an alternative workaround 🤦‍♂

